PR #100, opened 2023-01-26
allow pattern matching segment variables to take predicates

sritchie:
- #100:

  - Implements predicate support for `segment`, `entire-segment` and
    `reverse-segment` in `emmy.pattern.match`. This support bubbles up to forms
    in rules like `(?? x pred1 pred2)`.

  - Removes the `:emmy.pattern/ignored-restriction` linter keyword, and all
    clj-kondo code warning that restrictions aren't supported on segment binding
    forms.

### DIFF
--- a/src/emmy/pattern/match.cljc
+++ b/src/emmy/pattern/match.cljc
@@ -296,22 +296,25 @@
    (as-segment-matcher
     (fn segment-match [frame xs succeed]
       (let [xs (core/or xs [])]
-        (when (and (sequential? xs) (pred xs))
+        (when (sequential? xs)
           (if-let [binding (core/and
                             (core/not (s/wildcard? sym))
                             (frame sym))]
-            (let [binding-count (count binding)]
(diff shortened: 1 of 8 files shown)


src/emmy/pattern/match.cljc
```diff
@@ -500,7 +516,7 @@
    (let [match (pattern->combinators pattern)
          success (fn [frame]
                    (when-let [m (pred frame)]
-                     (when (and m (not (failed? m)))
+                     (when (core/and m (not (failed? m)))
```
sritchie, line 519: whoops!

sritchie:
@light-matters finally, here we go!

light-matters:
@sritchie Great stuff! Apologies for being so unproductive, time in the day etc. :(. Now that I have the Wolfram bridge working with my existing research simulations though then I'm feeling the pull back to 'utils (emmy?!).

PR #102, opened 2023-02-12
Add stream-integrator to ode

littleredcomputer:


littleredcomputer:
nb: you are also welcome to review the changes in odex-js

.gitignore
```diff
@@ -15,5 +15,9 @@ node_modules
 **.shadow-cljs
 package-lock.json
 public/
-.clj-kondo/.cache
-.clerk
\ No newline at end of file
+/.clj-kondo/
```
sritchie, line 18: the other stuff in this directory gets copied in by clj-kondo and gets used to lint macros etc in dependencies; are you thinking that each user should run that command on their own?

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+   segments."
+  [f' x0 y0 {:keys [epsilon] :or {epsilon default-epsilon}}]
+  (let [dimension (count y0)]
+    #?(:clj
+       (let [gbs               (GraggBulirschStoerIntegrator. 0. 1. (double epsilon) (double epsilon))
+             ode               (ExpandableStatefulODE.
+                                (reify FirstOrderDifferentialEquations
+                                  (computeDerivatives
+                                    [_ x y out]
+                                    (let [y' (f' x y)]
```
sritchie, line 57: adding a note for us to discuss — I was thinking while writing the animations that we might make the ODE interface require a precompiled function 

```clj
(f' x y out)
```

where `f'` is responsible for mutating the array. Then on the way in, we can wrap an `f'` that returns a clojure sequence:

```clj
(defn ->arr
  [^doubles arr ^Iterable xs]
  (let [it (.iterator xs)]
    (loop [i 0]
      (when (.hasNext it)
        (aset arr i (double (.next it)))
        (recur (unchecked-inc i))))))

(defn ->raw [f]
  (fn [x y out]
    (->arr out (f x y))))
```

I benchmarked a `->raw*` that used the `doseq` (which is how I initially wrote it too, before polling Clojurians Slack and getting the above solution), and found that `->arr` above is ~10-15x faster in my microbenchmark:

```clj
(defn ->arr
  [^doubles arr ^Iterable xs]
  (let [it (.iterator xs)]
    (loop [i 0]
      (when (.hasNext it)
        (aset arr i (double (.next it)))
        (recur (unchecked-inc i))))))

(defn ->raw [f]
  (fn [x y out]
    (->arr out (f x y))))

(defn ->raw* [f]
  (fn [x y ^doubles out]
    (let [y' (f x y)]
      (doseq [i (range 0 (alength out))]
        (aset out i (double (nth y' i)))))))

(def out (double-array 3))

(let [f1 (->raw  (fn [_ _] [1 2 3]))
      f2 (->raw* (fn [_ _] [1 2 3]))
      n  100000000]
  (time
   (dotimes [_ n]
     (f1 1 2 out)))
  (time
   (dotimes [_ n]
     (f2 1 2 out))))
;; Elapsed time: 1016.89225 msecs
;; Elapsed time: 16735.008792 msecs
```

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+   JavaScript, we expect the solver to provide a generator of solution
+   segments."
+  [f' x0 y0 {:keys [epsilon] :or {epsilon default-epsilon}}]
+  (let [dimension (count y0)]
+    #?(:clj
+       (let [gbs               (GraggBulirschStoerIntegrator. 0. 1. (double epsilon) (double epsilon))
+             ode               (ExpandableStatefulODE.
+                                (reify FirstOrderDifferentialEquations
+                                  (computeDerivatives
+                                    [_ x y out]
```
sritchie, line 56: I'm confusing myself, how were we ignoring `x` before??

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+   without accumulating unbounded state.
+   
+   The returned function may be called with no arguments to shut down
+   the integration, allowing for the final reclamation of its resources.
+
+   When the ODE solver is provided by Java, it may be necessary to 
+   use an auxiliary thread to enable this style of flow control.  If
+   JavaScript, we expect the solver to provide a generator of solution
+   segments."
+  [f' x0 y0 {:keys [epsilon] :or {epsilon default-epsilon}}]
```
sritchie, line 49: are we letting `y0` be a structure here? If so we can use `emmy.structure/dimension`.

src/emmy/numerical/ode.cljc
sritchie: The only other change I would suggest is that we allow the user to specify stateful parameters that can change. In scmutils, these come in via the first layer of the "state derivative"; I found that it worked really nicely to

- take a vector of parameters as an extra argument, like `(f' x y params out)`, if the function needs them
- Take an ATOM to supply the params, and internally call `(f' x y (.-state state-atom) out)`.

The reason to call `(.-state state-atom)` vs `@` aka `deref` is that accessing `-state` directly on the JS side will get around Reagent's "do I re-render" watcher, which checks for deref calls.

src/emmy/numerical/ode.cljc
littleredcomputer: Because Sussman bakes "time" into the 0th element of the state tuple, state-derivatives compute time by integrating $1\,dx$.

src/emmy/numerical/ode.cljc
littleredcomputer: Definitely worth thinking about. At the bottom, the solvers now both work with primitive arrays, and we could save an allocation by allowing the user of the primitive system to mutate an out vector instead of returning something. That's kind of a defect of the way I set up odex-js, though, since they are expected to return an array

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+                              (a/>!! step-requests true)
+                              (a/<!! solution-segments))
+               current-segment (atom (next-segment))]
+           (fn
+             ([]
+              (a/>!! step-requests false))
+             ([x]
+              (when (< x (:x0 @current-segment))
+                (throw (IllegalStateException. "Cannot use interpolation function in backwards direction")))
+              (while (> x (:x1 @current-segment))
```
sritchie, line 134: AWESOME!!

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+              ;; The `step-requests` channel sends `true` each time a new segment of 
+              ;; the solution is demanded; `false` is a signal that the consumer
+              ;; has no further need of them. When sending segments back, receiving
+              ;; false from the send operation indicates that the channel is closed.
+              ;; In either of these cases the integration should be shut down. 
+              ;;
+              ;; Normally, the use of an exception for flow control is frowned upon.
+              ;; Ideally, the CM3 integrator would look at the return value of the 
+              ;; step handler to decide whether or not to continue, but it doesn't.
+              ;; (You can install an EventHander, which watches an indicator function
```
sritchie, line 81: typo `EventHandler`

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+                                          x1 (.getCurrentTime interpolator)]
+                                      {:x0 x0
+                                       :x1 x1
+                                       :f  (fn [x]
+                                             (assert (and (>= x x0)
+                                                          (<= x x1))
+                                                     (format "ODE interpolation request %g out of range [%g, %g]"
+                                                             x x0 x1))
+                                             (.setInterpolatedTime interpolator x)
+                                             (into [] (.getInterpolatedState interpolator)))}))
```
sritchie, line 97: oh, yeah, I think I have the same feeling here, that `f'` should actually expect a double array, and should NOT keep a reference. Then we can wrap it to do this `(into [] ...)`.

The reason I want to do this is that I want to modify the `compile-fn` code (and did this in a branch already) to be able to generate `(aget y ...)` calls instead of destructuring or `nth` calls. This ends up being a good bit faster.

src/emmy/numerical/ode.cljc
```diff
@@ -2,96 +2,171 @@
...
+                (let [s (next-segment)]
+                  (reset! current-segment s)))
+              ((:f @current-segment) x)))))
+
+       :cljs
+       (let [jsf' (fn [x ys] (double-array (f' x ys)))
+             solver (o/Solver.
+                     jsf'
+                     dimension
+                     (clj->js {:absoluteTolerance epsilon
```
sritchie, line 144: not that it really matters for performance, but if you're writing a literal you can just do `#js {:key "value" ...}` to directly create the js array here, instead of `js->clj`.

src/emmy/numerical/ode.cljc
sritchie: @littleredcomputer that's true re: Odex!! but to play nice with Odex, we can allocate a single `out` array for return values, pass it into `f'` and then continually return the same `out` array to Odex.

.gitignore
littleredcomputer: What happened was I started using Calva (Clojure plugin for VS Code) and found the following uncommitted files, from git's perspective:
```bash
	.calva/
	.clj-kondo/babashka/
	.clj-kondo/org.mentat/
	.clj-kondo/rewrite-clj/
	.clj-kondo/taoensso/
	.cpcache/
	.lsp/
```
so I wanted to preserve the status quo unless we figure out what the value of checking in those things would be. I am guided by [this discussion](https://clojurians-log.clojureverse.org/clj-kondo/2022-03-18), which, while equivocal (as I read it), seems to indicate that at least I'm doing no harm

src/emmy/numerical/ode.cljc
littleredcomputer: I am convinced, so I will rework things to allow a preallocated array.

src/emmy/numerical/ode.cljc
littleredcomputer: I waffled on this. I can think of three different kinds of derivative.
- state derivatives (x is ignored, state is not flat)
- clojure (x matters, state is a seq of the customer's choosing, but probably `[...]`)
- native (x matters, state is a primitive array)
and my decision was that a good ODE library would be clojure-y. I suppose we can get away with it if we insist on having the user mutate the array given, so the input state f' receives is a primitive array but the user won't notice because it's a seq, same for the array that the user fills up. What I would lament slightly is not having the ability to write derivatives like `(x, [y0, y1]) => [-y1, y0]`

src/emmy/numerical/ode.cljc
littleredcomputer: neat thank you!

src/emmy/numerical/ode.cljc
littleredcomputer: Rereading what you have written carefully, let me see if I have it right.

- `stream-integrator` will insist on being called with a "raw function", i.e., one with the signature [x y out] where y and out are [D. 
- Users of the classic SICM interface will have their derivatives wrapped into raw form.
- Since the compiler is hidden inside the SICM-style interface, you will have the privilege of relying on this so the compiler can produce a derivative that fulfills the raw contract, skipping the wrapping when the compiler is used.
- I will change the expected function signature in ODEX to raw, so that we will lose no performance in the JS case. I'm making a 3.0.0 version, so I'm permitted this breaking change. 
- I could then add an option requesting the old calling convention 
- Or I could be "magic" and say if f' returns "something" then that's the value else the contents of `out`, so it magically works with old functions and new. But I have talked myself out of this

littleredcomputer: On the subject of `->arr`: Currently, what goes into the fixed array after the state derivative is called is the result of flattening the state increment. `flatten` produces a LazySeq, which, unfortunately, is not considered `Iterable` in cljs, so that code snippet doesn't work.

Further study revealed that the long pole time wise is actually the `flatten` itself, about 10x, which dwarfs the difference between `doseq` and `->arr`. 

Casting about for a solution that was better than the naive approach while being portable to cljs without hacking, I decided to try `clojure.core.reducers` and the good news is that the flatten in the reducers library is much more performant in the microbenchmark than `clojure.core/flatten`. This coupled with an `aset` in the reducing step seems like a fair compromise

```clojure
(require '[clojure.core.reducers :as r])

(defn flatten-into-primitive-array [a s]
  (let [ix (atom -1)
        asetter aset-double]
    (r/reduce (fn [a t]
                (asetter a (swap! ix unchecked-inc) t)
                a)
              a
              (r/flatten s))))

(let [a (double-array 7)
      s [1. [2. 3. 4.] [5. 6. 7]]
      n 10000000]
  (time
   (dotimes [_ n]
     (->arr a (flatten s))))
  (time
   (dotimes [_ n]
     (flatten-into-primitive-array a s))))

"Elapsed time: 21511.873042 msecs"
"Elapsed time: 6022.566583 msecs"
```

so while your solution was down to ~600msec, it was exclusive of flatten. For this variant of the microbenchmark, we are flattening something which looks very much like a Sussmanian Derivative

src/emmy/numerical/ode.cljc
littleredcomputer: No. `stream-integrator` by design offers a `double-array -> double-array` interface, leaving the SICM layer to adapt state-derivatives in this form.

src/emmy/numerical/ode.cljc
littleredcomputer: done (and you might even get the compile-fn code to generate `(aset-double y ...)` on the way out

src/emmy/numerical/ode.cljc
littleredcomputer: fixed

src/emmy/numerical/ode.cljc
littleredcomputer: fixed

src/emmy/numerical/ode.cljc
littleredcomputer: this has been implemented and further simplifies ode.clj, no longer requiring different versions of `state->array` for clj and class

littleredcomputer: Returning now to the subject of atoms, wherein you request another parameter in the derivative to hold the time-varying values, would it not be possible to just close over the atoms you need without involving the ODE interface?

sritchie:
@littleredcomputer review inbound tomorrow... I finished up my six template projects today and it's full steam ahead on Emmy now!

src/emmy/numerical/ode.cljc
```diff
@@ -249,11 +289,13 @@
...
-                    (swap! out conj state))]
-    (I initial-state dt t1 {:compile? true
-                            :epsilon 1e-6
-                            :observe collector})
-    @out))
+  (let [opts (integration-opts state-derivative
+                               state-derivative-args
+                               initial-state
+                               {})
+        f (:integrator opts)
```
sritchie, line 296: this is all really nice!

src/emmy/numerical/ode.cljc
```diff
@@ -249,11 +289,13 @@
...
+        v (mapv f (for [x (range 0 (+ t1 dt) dt)
```
sritchie, line 297: style wise you could do this in a `try, finally` , but treat that as me musing, not a request

src/emmy/numerical/ode.cljc
sritchie: yes, it would for sure, if you tweak `compile-state-fn` to expect an atom and then internally deref it... we can look at these styles when we get the animation example up and start hacking on that.

sritchie:
This is excellent work... merge away, I can't wait to use this on our animations!

src/emmy/numerical/ode.cljc
littleredcomputer: oh, I see what you mean. I'll fix it. That or a `with-stream-integrator` macro.

littleredcomputer:
Thank you for the review. A lot of glue melted away, always a good sign.

PR #104, opened 2023-02-28
Fix odex version, type hint

sritchie:


sritchie:
I think these all made it in already.

PR #105, opened 2023-02-28
Fn speedup [still porting]

sritchie:


sritchie:
This might be useful:

```clj
(defn ^:no-doc array-bindings
  "
  - `include?`: fn (or a set!) that returns true of the symbol it's passed is
    used inside of the function body (and needs to be bound), false otherwise.
  - `array-sym`: symbol that stands for the array version of `inputs`.
  - `inputs`: the structure of symbols that we're converting into a primitive
    array during compilation."
  [include? array-sym inputs]
  (into [] (comp (map-indexed
                  (fn [i v]
                    (when (include? v)
                      [v `(aget ~array-sym ~i)])))
                 cat)
        (flatten inputs)))

(comment
  ;; For example:
  (= '[t (clojure.core/aget arr 0)
       thetadot (clojure.core/aget arr 2)]
     (array-bindings '#{+ t thetadot}
                     'arr
                     '[t theta thetadot])))
```

2. To gensym a variable that won't throw reflection warnings on `aget` and `aset`, use this:

```clj
(with-meta (gensym) {:tag 'doubles})
```

3. This was a snippet to generate the output `aset` calls:

```clj
(defn ^:no-doc ->output [body output]
  (if (= `let (first body))
    `(~@(take 2 body)
      ~(->output (nth body 2) output))
    `(doto ~output
       ~@(map-indexed (fn [i x]
                        `(aset ~i ~x))
                      (flatten-structs body)))))

(comment
  (= '(clojure.core/doto output
        (clojure.core/aset 0 t)
        (clojure.core/aset 1 x0)
        (clojure.core/aset 2 x1)
        (clojure.core/aset 3 v0)
        (clojure.core/aset 4 v1))
     (->output
      '(up t (up x0 x1) (up v0 v1))
      'output)))
```

PR #107, opened 2023-03-03
native JS function compilation

littleredcomputer:
- move CSE to its own namespace to avoid the circular dependency compile->render->compile

- refactor JS rendering to allow compiler to use it

- adjust meaning of :native and :source compilation modes now you get what's compatible with your execution environment but you can also ask for a specific language, allowing tests to be bilingual

### DIFF
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -151,7 +151,7 @@
 ;; Armed with the above compiler optimization we can move on to the actual
 ;; compilation step.
 ;;
-;; This library provides two compilation modes:
+;; This library provides several compilation modes:
 ;;
 ;; - Native compilation via `eval` in Clojure, transpiling in JavaScript
 ;; - interpreted compilation via [SCI](https://github.com/borkdude/sci), the
@@ -161,7 +161,7 @@
 ;; - just asking for :source will provide the source in whichever environment
 ;;   you're using
 ;;
-;; We default :native for performance.
+;; We default to :native for performance.
 
 (def ^{:dynamic true
        :no-doc true}
@@ -197,10 +197,10 @@
   #?(:cljs (set! *mode* mode)
      :clj  (alter-var-root #'*mode* (constantly mode))))
 
-;; Native compilation works on the JVM, and on ClojureScript if you're running
-;; in a self-hosted CLJS environment. Enable this mode by wrapping your call in
+;; Native compilation works on the JVM and in ClojureScript. Enable this mode
+;; by wrapping your call in
 ;;
-;; `(binding [*mode* :native] ,,,)`
+;; (binding [*mode* :native] ,,,)`
 ;;
 ;; ## State Functions
 ;;
@@ -245,7 +245,7 @@
       [state (into [] params)]
       [state])))
 
-;; The following two functions compile state functions in either native or SCI
+;; The following functions compile state functions in either native or SCI
 ;; mode. The primary difference is that native compilation requires us to
 ;; explicitly replace all instances of symbols from `compiled-fn-whitelist`
 ;; above with actual functions.
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -302,24 +302,22 @@
 (defn- compile->clj
   "Returns Clojure source for a function that implements `body`, given:
 
-  - `params`: a seq of symbols equal in count to the original state function's
-    args
-  - `state-model`: a sequence of variables representing the structure of the
-    nested function returned by the state function
-  - `body`: a function body making use of any symbol in the args above"
+  - `argv`: a vector of symbols to serve as the function's arguments.
+    The vector may be nested for sequence destructuring.
+
+  - `body`: a function body making use of any symbol in argv above"
   [argv body]
   (let [body (w/postwalk-replace sym->resolved-form body)]
     `(fn [~@argv] ~body)))
 
 (defn- compile->js
   "Returns an array containing JavaScript source for a function that implements
-  `body` in a form suitable for the arguments of the Function constructor, given:
+  `body` in a form suitable for application of the Function constructor, given:
 
-  - `params`: a seq of symbols equal in count to the original state function's
-    args
-  - `state-model`: a sequence of variables representing the structure of the
-    nested function returned by the state function
-  - `body`: a function body making use of any symbol in the args above"
+  - `argv`: a vector of symbols to serve as the function's arguments.
+    The vector may be nested for sequence destructuring.
+
+  - `body`: a function body making use of any symbol in argv above"
   [argv body]
   (let [argv        (mapv commafy-arglist argv)
         js          (render/->JavaScript* body {})
@@ -334,11 +332,10 @@
 (defn- compile-native
   "Returns a natively-evaluated function that implements `body`, given:
 
-  - `params`: a seq of symbols equal in count to the original state function's
-    args
-  - `state-model`: a sequence of variables representing the structure of the
-    nested function returned by the state function
-  - `body`: a function body making use of any symbol in the args above"
+  - `argv`: a vector of symbols to serve as the function's arguments.
+    The vector may be nested for sequence destructuring.
+
+  - `body`: a function body making use of any symbol in argv above"
   [argv body]
   #?(:clj (eval (compile->clj argv body))
      :cljs (comp js->clj (apply js/Function (compile->js argv body)))))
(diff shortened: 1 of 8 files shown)


src/emmy/expression/compile.cljc
```diff
@@ -124,232 +123,6 @@
...
-   (letfn [(callback [new-expression bindings]
-             (let [n-bindings (count bindings)]
-               (if (pos? n-bindings)
-                 (let [binding-vec (into [] cat bindings)]
-                   (log/info
-                    (format "common subexpression elimination: %d expressions" n-bindings))
-                   `(let ~binding-vec
-                      ~new-expression))
-                 new-expression)))]
-     (extract-common-subexpressions expr callback opts))))
```
sritchie, line 351: I wonder if there is some way (not now, down the road) to use this feature in the TeX renderer? Is there some standard way to describe local bindings in TeX?

src/emmy/expression/compile.cljc
```diff
@@ -421,10 +197,10 @@
   #?(:cljs (set! *mode* mode)
      :clj  (alter-var-root #'*mode* (constantly mode))))
 
-;; Native compilation works on the JVM, and on ClojureScript if you're running
-;; in a self-hosted CLJS environment. Enable this mode by wrapping your call in
+;; Native compilation works on the JVM and in ClojureScript. Enable this mode
+;; by wrapping your call in
 ;;
-;; `(binding [*mode* :native] ,,,)`
+;; (binding [*mode* :native] ,,,)`
```
sritchie, line 203: I think this needs its leading backtick back

src/emmy/expression/compile.cljc
```diff
@@ -463,13 +239,13 @@
                        :or {flatten? true
                             generic-params? true}}]
   (let [state (into [] (if flatten?
-                         (into [] (flatten state-model))
```
sritchie, line 466: good catch on my double into!

src/emmy/expression/compile.cljc
```diff
@@ -491,6 +267,31 @@
                  {'up struct/up
                   'down struct/down}}}))
 
+(defn- commafy-arglist
+  "Since JavaScript likes commas between array elements:
+
+   [a [b c [d e] f]] -> \"[a, b, c, d, e, f]\"       if flatten?
+   [a [b c [d e] f]] -> \"[a, [b, c, [d, e], f]]\"   if not flatten?
```
sritchie, line 274: did you mean to add a flatten? keyword arg?

src/emmy/expression/compile.cljc
```diff
@@ -499,31 +300,51 @@
...
   (sci/eval-form (sci/fork sci-context) f-form))
 
-(defn- compile-state->source
-  "Returns a natively-evaluated Clojure function that implements `body`, given:
+(defn- compile->clj
+  "Returns Clojure source for a function that implements `body`, given:
 
   - `params`: a seq of symbols equal in count to the original state function's
     args
   - `state-model`: a sequence of variables representing the structure of the
```
sritchie, line 308: nice simplification, let's update this part of the docstring too.

src/emmy/expression/compile.cljc
```diff
@@ -499,31 +300,51 @@
...
+  - `params`: a seq of symbols equal in count to the original state function's
+    args
+  - `state-model`: a sequence of variables representing the structure of the
+    nested function returned by the state function
+  - `body`: a function body making use of any symbol in the args above"
+  [argv body]
+  (let [argv        (mapv commafy-arglist argv)
+        js          (render/->JavaScript* body)
+        cs-bindings (s/join (for [[var val] (:vars js)]
+                              (str "  const " var " = " val ";\n")))
```
sritchie, line 328: is there any benefit to using `let` here, or is it the other way around, that `var` has higher performance but potentially less safety?

src/emmy/expression/compile.cljc
```diff
@@ -499,31 +300,51 @@
...
   - `params`: a seq of symbols equal in count to the original state function's
     args
   - `state-model`: a sequence of variables representing the structure of the
     nested function returned by the state function
   - `body`: a function body making use of any symbol in the args above"
-  [params state-model body opts]
-  (eval
-   (compile-state->source params state-model body opts)))
+  [argv body]
+  #?(:clj (eval (compile->clj argv body))
```
sritchie, line 344: rather than compose with `js->clj`, what do you think of making an explicit call inside the ->JavaScript` function to a clojure vector constructor that we export? That way we could use this same function to compile functions that mutate the output vector, or return native js arrays by design. Thoughts?

src/emmy/expression/compile.cljc
```diff
@@ -532,9 +353,9 @@
   - `state-model`: a sequence of variables representing the structure of the
```
sritchie, line 353: I'm quite happy that this seems to be unifying the different compile functions, between the compile-state-fn and compile-fn.

Do you think it makes sense to have a mode that compiles something that looks like a state derivative into a function of the parameters that returns a function of the local tuple, instead of flattening both levels into one function? Not for this PR, but I wanted to throw that into the design soup.

src/emmy/expression/render.cljc
```diff
@@ -661,27 +644,73 @@
...
+
+  Otherwise, it is interpreted as the sequence of parameters itself. If not
+  specified, the default behavior is `sort`.
+
+  If `:deterministic? true` is supplied, the function will assign variables by
+  sorting the string representations of each term before assignment. Otherwise,
+  the nondeterministic order of hash maps inside this function won't guarantee a
+  consistent variable naming convention in the returned function. For tests, set
+  `:deterministic? true`."
+  [x & {:keys [parameter-order]
```
sritchie, line 680: Okay, I remember why I moved this function over. It seemed to me that this function was doing something sympy-like by taking an expression and trying to treat it as a function of the unspecified arguments.

In a similar way to the compile functions, it felt like it might be more consistent to have ->JavaScript take a function, and then generate the arguments and pass them into the fn instead of trying to infer and pull them out (which runs into argument order trouble, if you want to do something other than manually inspect the source before compiling.)

src/emmy/expression/render.cljc
```diff
@@ -661,27 +644,73 @@
...
+        :or   {parameter-order sort}
+        :as opts}]
+  (let [{:keys [params vars value]} (->JavaScript* x opts)
+        ordered-params (if (fn? parameter-order)
+                         (parameter-order params)
+                         parameter-order)]
+    (with-out-str
```
sritchie, line 687: I think this approach of with-out-str is fine for now, but I did hit some bugs a while ago where a println somewhere deep inside of emmy was triggering and getting into the string. No need to change it now, but I wanted to get that down somewhere!

sritchie (changes requested):
This is looking great. Would you mind adding the notes from the PR to the CHANGELOG?

src/emmy/expression/compile.cljc
littleredcomputer: It could be done. In fact, it might even be cool. Something like (to borrow a Haskellism):
$$
\bf{let}
\qquad A_1 := x+y\cr
\qquad A_2 := \sin(x)\cr
\bf{in}
\qquad A_1^2 + A_2
$$
I don't know if that would actually render. I was sort of hoping GitHub would try to

src/emmy/expression/compile.cljc
littleredcomputer: done

src/emmy/expression/compile.cljc
littleredcomputer: no, this came out in the wash (fixed doc)

src/emmy/expression/compile.cljc
littleredcomputer: The [experts say](https://stackoverflow.com/questions/40070631/v8-javascript-performance-implications-of-const-let-and-var) it doesn't matter. I'm on the fence. Having been raised on const-correctness in ANSI C, I like to use const when possible, but if you write JS in modern style, practically everything is const except loop indices.

src/emmy/expression/compile.cljc
littleredcomputer: I put this in because I was getting a weird error of "IReduce.-reduce" not being present which I tracked down to the JS-nature of the returned array. This caused me to conclude that a JS array looked innocent but could fail to work in some distant context. Supplying a cli vector compatible constructor would do the trick, at the cost of the compiled code not looking like "real" JS. 

But you point out that this is a copy in performance critical code ... I'd rather "fix" this with an `aset` output convention than inject the constructor. I have personally benefited from the ability to paste the output into browser code, so I want to keep it.

src/emmy/expression/render.cljc
littleredcomputer: We found the same thing. This is why I separated the functions via the map return value: nobody in compile-land looks at the inferred argument list, which is only used by the "printing" version, which I now agree is bogus (or at least that you have to manually deal with the function wrapping).

One way out of this would be to just have the ("legacy") `->JavaScript` just print the function body and leave the construction of the execution context up to the consumer (as compile.cljs does) instead of printing something that probably won't work

src/emmy/expression/render.cljc
littleredcomputer: Good point. I heard a slight discordant note in the soundtrack as I used this feature and now repent of it. I'll fix it

littleredcomputer:
CHANGELOG updated

littleredcomputer:
ptal

sritchie:
LGTM, once we get this merged I'll test it out with the animation code. So good!!

PR #109, opened 2023-03-07
unify JavaScript with other kinds of compilation

littleredcomputer:
- `->JavaScript` now works on S-expressions only, like other renderers and does not emit function wrapping

- more principled approach for unary negation

- unified CSE path for Clojure and JS compilation


littleredcomputer:
at this point the Clojure part works, but there are some divergences when the JS is compiled by Clojurescript. Some more tweaks will be necessary to get everything consistent

src/emmy/expression/compile.cljc
```diff
@@ -369,6 +373,25 @@
               (gensym-fn 'y)))]
     (rec state)))
 
+(defn- cse
+  "Invokes extract-common-subexpressions on `x`.
```
sritchie, line 377: If you write this as [[emmy.expression.cse/extract-common-subexpressions]] it will turn into a hyperlink in our cljdoc. Annoyingly this syntax won't work for vars in clojure.core or anywhere else outside the project, nor will it follow aliases in the namespace... [[var-inside-namespace]] will work without prefix, though.

src/emmy/expression/compile.cljc
```diff
@@ -369,6 +373,25 @@
...
+   The results of the elimination are presented in a map with the keys
+   `:bindings` and `body`.  If no suitable subexpressions were found,
+   the binding list will be empty and body will be unchanged.
+
+   "
+  [x opts]
+  (extract-common-subexpressions
+   x
+   (fn [body bindings]
+     ;; TODO: consistency demands that we do apply-numeric-ops on the
```
sritchie, line 388: yup, and it makes sense to do it here instead of before subexpression elimination so we don't repeat all of the work, yeah?

src/emmy/expression/compile.cljc
```diff
@@ -369,6 +373,25 @@
...
+     ;; intermediate bindings, but that will shake up a lot of unit
+     ;; tests so I'm temporarily postponing this.
+     {:bindings bindings #_(into {} (for [[k v] bindings] [k (apply-numeric-ops v)]))
```
sritchie, line 391: oh, let's be careful to actually NOT put these into a map, since that might mix up the ordering, which definitely matters here. I think these bindings need to stay a sequence of pairs. Is that how they come into the callback?

src/emmy/expression/compile.cljc
```diff
@@ -481,8 +505,7 @@
          body     (-> (apply f args)
                       (g/simplify)
                       (v/freeze)
-                      (cse-form)
-                      (apply-numeric-ops))
+                      (cse {}))
```
sritchie, line 508: You might also give `cse` a zero-arity version that supplies this default empty options map

src/emmy/expression/render.cljc
```diff
@@ -39,9 +36,12 @@
   simplifier negates by wrapping with `(* -1 ...)`. For rendering, we prefer to
   use a unary minus."}
   rewrite-negation
+
   (R/ruleset
    (* -1 ?x) => (u- ?x)
-   (* -1 ??x) => (u- (* ??x))))
+   (* -1.0 ?x) => (u- ?x)
+   (* -1 ??x) => (u- (* ??x))
```
sritchie, line 43: You could also write this pattern as `(* (? _ #{-1.0 -1}) ??x)`

src/emmy/expression/render.cljc
```diff
@@ -163,22 +164,47 @@
...
-                               (if-not (and (string? (first a')) (= (first (first a')) \-))
-                                 (print sep)
-                                 (print " "))
-                               (recur a')))))))
+
+                       (cond
+                         (= (count args) 1)
+                         (render-unary-node op (first args) upper-op)
+
+                         ;; Special handling for + over unary -. The first term
```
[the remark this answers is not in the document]
sritchie: I'm asking @borkdude in Slack if he has thoughts on let vs const here for generated bindings, as I (like you) would have reached for `const`.

src/emmy/expression/render.cljc
```diff
@@ -163,22 +164,47 @@
...
+                               terms (cons (cond
+                                             (= (first t) "+")
+                                             (subvec t 1)
+
+                                             (= (first t) "-")
+                                             (assoc t 0 "- ")
+
+                                             :else t)
+                                           (for [[pm t] terms] [(str " " pm " ") t]))]
+                           (transduce cat str terms))
```
sritchie, line 199: I love me some transducers...

test/emmy/examples/double_pendulum_test.cljc
```diff
@@ -60,81 +70,95 @@
...
-                "  var _000e = Math.cos(_000a);\n"
-                "  var _0010 = Math.sin(_000a);\n"
-                "  var _0011 = Math.pow(_0010, 2);\n"
-                "  return [1, [thetadot, phidot], [(l1 * m2 * _0006 * _000e * _0010 + l2 * m2 * _0005 * _0010 + g * m2 * _0008 * _000e - g * m1 * _0009 - g * m2 * _0009) / (l1 * m2 * _0011 + l1 * m1), (- l2 * m2 * _0005 * _000e * _0010 - l1 * m1 * _0006 * _0010 - l1 * m2 * _0006 * _0010 + g * m1 * _0009 * _000e + g * m2 * _0009 * _000e - g * m1 * _0008 - g * m2 * _0008) / (l2 * m2 * _0011 + l2 * m1)]];\n"
-                "}")
-           (e/->JavaScript eq
-                           :parameter-order '[t theta phi thetadot phidot]
-                           :deterministic? true))))
+  (is (=
+       ["[y01, [y02, y03], [y04, y05]]"
```
sritchie, line 74: does this end up having different performance than manually generating the bindings, or are they equivalent in performance? It would be nice if they were equivalent...

test/emmy/examples/driven_pendulum_test.cljc
```diff
@@ -5,10 +5,19 @@
   (:require [clojure.test :refer [deftest is use-fixtures]]
             [emmy.env :as e :refer [up /]]
             [emmy.examples.driven-pendulum :as driven]
+            [emmy.expression.compile :refer [compile-state-fn*]]
             [emmy.simplify :refer [hermetic-simplify-fixture]]))
 
 (use-fixtures :each hermetic-simplify-fixture)
 
+(defn- gensym-fn
```
sritchie, line 13: you think it's worth putting this somewhere common? Folks are probably going to want something like this for the same reasons we want it in the tests, and I can't remember if we provide it already.

test/emmy/expression/cse_test.cljc
```diff
@@ -60,29 +60,3 @@
     (is (= [slimmed expected-subs]
            (c/extract-common-subexpressions expr vector opts))
         "subexpressions are again extracted in order.")))
-
-(def letsym
```
sritchie, line 64: In the future, I now know that we can do this by writing `let, which will resolve correctly to the namespaced symbol.

test/emmy/sicm/ch3_test.cljc
```diff
@@ -17,6 +18,14 @@
 (def simplify
   (comp e/freeze e/simplify))
 
+(defn- gensym-fn
```
sritchie, line 21: here is our friend again!

src/emmy/expression/render.cljc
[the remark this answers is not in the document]
sritchie: "I don't know if there's a performance cost to either. I would assume engines to optimize both"

sritchie (changes requested):
Looking good, the tests looks nice and clean now without the function body.

Part of me still wonders if we could get away with clojure code as our source representation, instead of the map of bindings and code body living separately.

We could put handlers for `let` and `fn` into `->JavaScript` and get the same results, I think, right? That would also let us compile higher order functions etc by only worrying about how we were going to build the function as a Clojure form.

Let me know what you think, or what downsides I'm missing here...

littleredcomputer:
> Part of me still wonders if we could get away with clojure code as our source representation, instead of the map of bindings and code body living separately...We could put handlers for `let` and `fn` into `->JavaScript`

We can in fact do this and maybe we should. I still believe, though, that the job of the compiler is to produce something which can be fed directly to the environment's code constructor which we have decided are eval and js/Function respectively, so the JS should come out in array form, suitable for that application. But should that be render's job? If we were willing to use JS eval, my position would be weaker

sritchie:
@littleredcomputer oh that's a good point. I will stand by my comment about internal bindings with let etc, but I agree with you now about the function boundary being a different animal, and higher order functions being trickier in a world without eval.

Let's let this continue simmering!

test/emmy/examples/driven_pendulum_test.cljc
littleredcomputer: Getting this far has made me want to crack the whip on this gensym business. Stay tuned.

(my current thought is something like this: there's no need to economize symbol generators. Each compile job can start out with a fresh one, that counts from 1, and can have various prefixes, so you'll get `["[y1, y2]" "[p3 p4]" " const _5 = ...] ` (modulo zero padding for the sort order, but we can get away with fewer digits since each function starts from scratch). If we thought 4 digits were enough, then they're short enough that the test expected outputs won't be horrible to look at, and the user will not need the ability to inject a generator because the compile outputs will always be stable).

test/emmy/expression/cse_test.cljc
littleredcomputer: gulp. I didn't even think of that, but that's indeed how it works!

test/emmy/sicm/ch3_test.cljc
littleredcomputer: for the present :imp:

src/emmy/expression/compile.cljc
littleredcomputer: 😮 you're right. Thank you. I forgot about this

src/emmy/expression/compile.cljc
littleredcomputer: done

src/emmy/expression/compile.cljc
littleredcomputer: done

src/emmy/expression/compile.cljc
littleredcomputer: done

test/emmy/examples/driven_pendulum_test.cljc
littleredcomputer: I have made it so that `monotonic-symbol-generator` is the only one left, but it's in the `analyze` namespace, which is not someplace we want our users to poke around. Open to suggestions as to where to move it (could put it in compile, for example)

test/emmy/sicm/ch3_test.cljc
littleredcomputer: done

src/emmy/expression/render.cljc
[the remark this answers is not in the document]
littleredcomputer: @borkdude 's output looks good. My personal feeling is that putting a bit of polish into the code generation is worth it from the point of view of a renderer like `->TeX`, where we'll really be judged for getting it wrong, and let JS leverage that effort

src/emmy/expression/render.cljc
littleredcomputer: done

src/emmy/expression/analyze.cljc
```diff
@@ -81,25 +82,53 @@
...
+   (take 5 (repeatedly j))
+   ;; (a000 a001 a002 a003 a004)
+   ;; (x005 x006 x007 x008 x009)
+   ```
+   "
+  ([]
+   (monotonic-symbol-generator 4 "_"))
+  ([size]
+   (monotonic-symbol-generator size "_"))
+  ([size prefix]
```
sritchie, line 120: Should we assert that the prefix doesn't start with a number (making sure that we get valid symbols out), or is this too pedantic?

src/emmy/expression/compile.cljc
```diff
@@ -136,14 +137,19 @@
...
-           (let [[f & xs] expr]
-             (if-let [m (and (every? number? xs)
-                             (numeric-whitelist f))]
-               (u/double (apply (:f m) xs))
-               expr))
-           :else expr))
+     (cond
+       ;; Native integers in Clojure are excused from promotion to double,
+       ;; as we don't want every `1` to become `1.0` in output formats
+       #?@(:clj [(or (instance? Long expr)
```
sritchie, line 143: This still makes me nervous... I can live with the `(double 1)` idea too!

src/emmy/expression/render.cljc
```diff
@@ -39,9 +26,12 @@
   simplifier negates by wrapping with `(* -1 ...)`. For rendering, we prefer to
   use a unary minus."}
   rewrite-negation
+
   (R/ruleset
-   (* -1 ?x) => (u- ?x)
-   (* -1 ??x) => (u- (* ??x))))
+   (* (? _ #{-1 -1.0}) ?x)               => (u- ?x)
```
sritchie, line 31: I hate to add one more literal but do we need `1N` here too?

src/emmy/expression/render.cljc
```diff
@@ -59,6 +49,16 @@
                (v/integral? denom))
           (str num "/" denom))))
 
+(defn- kebab->snake
+  "'foo-bar becomes 'foo_bar. Some other special characters are replaced by strings."
+  [s]
+  (-> (str s)
+      (s/replace #"-" "_")
```
sritchie, line 56: could we use `clojure.core/munge` here? https://clojuredocs.org/clojure.core/munge

test/emmy/examples/double_pendulum_test.cljc
```diff
@@ -60,81 +63,98 @@
...
+                "  const _46 = Math.cos(_36);\n"
+                "  const _50 = Math.sin(_36);\n"
+                "  const _55 = Math.pow(_46, 2);\n"
+                "  const _56 = Math.pow(_46, 4);\n"
+                "  const _57 = Math.pow(_50, 2);\n"
+                "  const _58 = - 2.0 * _18 * _20 * _23 * _55;\n"
+                "  const _59 = 2.0 * _18 * _20 * m_1 * m_2 * _57;\n"
+                "  const _60 = _18 * _20 * _23 * _56;\n"
+                "  const _62 = _60 + _59 + _58 + _32 + _34;\n"
+
```
sritchie, line 148: not sure if you wanted these line breaks...

test/emmy/expression/compile_test.cljc
```diff
@@ -157,15 +153,15 @@
             (with-redefs [gensym (fn
                                    ([] (clojure.core/gensym))
                                    ([x] x))]
-              (is (= #?(:clj `(fn [~'x] (+ (* -1.0 ~'x) 28.0))
+              (is (= #?(:clj `(fn [~'x] (+ (* -1 ~'x) 28.0))
```
sritchie, line 156: I am still worried here... see this:

```clj
(set! *unchecked-math* :warn-on-boxed)

;; The first two emit warnings, the next two do not.
(let [f1 (fn [x] (+ (* -1 x) 28.0))
      f2 (fn [x] (+ (* -1.0 x) 28.0))
      f3 (fn [^double x] (+ (* -1.0 x) 28.0))
      f4 (fn [^double x] (+ (* (double -1) x) 28.0))
      n 10000000000]  
  (time (dotimes [_ n]
          (f1 5)))
  (time (dotimes [_ n]
          (f2 5)))
  (time (dotimes [_ n]
          (f3 5)))
  (time (dotimes [_ n]
          (f4 5))))
```

Maybe this leads to a different set of suggestions....

- How about a compromise of `(double -1)`, if we really don't like `.0`?
- I think we want to compose our symbol generator, for anything that is going to represent a double, with

```clj
(fn [sym] (with-meta sym {:tag 'double}))
```

So that we can properly hint to the compiler that the variables are double-valued. I think `(set! *unchecked-math* :warn-on-boxed)` should stay quiet if we do that, and squawk if we don't.

Thoughts?

test/emmy/expression/cse_test.cljc
```diff
@@ -50,7 +44,7 @@
   (let [expr '(+ (sin x) (expt (sin x) 2)
                  (cos x) (sqrt (cos x)))
         opts {:deterministic? true
-              :symbol-generator (make-generator "K")}
+              :gensym-fn (a/monotonic-symbol-generator 1 "K")}
```
sritchie, line 47: the tests look way better with the 1 and 2 length symbols...

src/emmy/expression/analyze.cljc
```diff
@@ -81,25 +82,53 @@
...
-  "Returns a function which generates a sequence of symbols with the given
-  `prefix` with the property that later symbols will sort after earlier symbols.
-
-  This is important for the stability of the simplifier. (If we just used
-  `clojure.core/gensym`, then a temporary symbol like `G__1000` will sort
-  earlier than `G__999`. This will trigger errors at unpredictable times,
-  whenever `clojure.core/gensym` returns two symbols that cross an
-  order-of-magnitude boundary.)"
-  [prefix]
-  (let [count (atom -1)]
```
sritchie, line 93: IIRC this `-1` was why the old version had symbols starting at 0... but probably not worth changing and making more work to migrate the tests, when you've done this so many times!

src/emmy/expression/render.cljc
```diff
@@ -6,25 +6,12 @@
...
-              (format "%s%04x" p (swap! i inc))
-
-              :cljs
-              (let [suffix (-> (swap! i inc)
-                               (.toString 16)
-                               (.padStart 4 "0"))]
-                (str p suffix)))))))
+            [emmy.value :as v]))
 
 (def ^{:private true
```
sritchie, line 16: once we get this merged I am going to move the docstrings down below the symbol, like in #110 . I seriously thought that was not a thing, leading to the 100s of `:doc` forms... whoops!

sritchie (changes requested):
Few more comments around performance, we're close here!! This is excellent work, thank you for all the care and attention here.

src/emmy/expression/analyze.cljc
littleredcomputer: I thought about this and there were just fewer tests to hack on with it done this way.

test/emmy/expression/compile_test.cljc
littleredcomputer: I now feel that it's not the job of this change to tamper with this. If you're happy with everything else, I will reintroduce the double cast of int/long and fix all the unit test expected outputs. For those that differ in output based on the runtime environment, I'll use `#?`.

src/emmy/expression/render.cljc
littleredcomputer: That's a great find. I like it (it does what I did and many other things) but oddly lets things like `(` through, but it would take effort to get that into a symbol name

src/emmy/expression/render.cljc
littleredcomputer: we're ok:
```clojure
user=> (#{1 1.0} 1N)
1
```

test/emmy/expression/compile_test.cljc
sritchie: Okay, let's do that - I've made the mistake of using these PR discussions to mix concrete feedback on the issue at hand with wide ranging musing. I agree, let's back this out now and then, armed with some animation etc examples, take another pass to see if there are performance wins we need to handle or if the JVM / browser are in fact smart enough that we get it all for free.

src/emmy/expression/compile.cljc
```diff
@@ -318,27 +324,25 @@
...
   "Returns a natively-evaluated function that implements `body`, given:
 
   - `argv`: a vector of symbols to serve as the function's arguments.
     The vector may be nested for sequence destructuring.
 
-  - `body`: a function body making use of any symbol in argv above"
-  [argv body]
-  #?(:clj (eval (compile->clj argv body))
-     :cljs (comp js->clj (apply js/Function (compile->js argv body)))))
+  - `body`: a map containing `bindings` and `body` keys"
```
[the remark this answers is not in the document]
littleredcomputer: done

src/emmy/expression/compile.cljc
littleredcomputer: done

PR #115, opened 2023-03-13
unify compile-fn and compile-state-fn; primitive mode for compiled fns

littleredcomputer:
From the CHANGELOG:

- #115:

  This PR introduces significant upgrades to the functions in
  `emmy.expression.compile`. `compile-state-fn` and `compile-fn` now share the
  same code. Expect some more shifts here as we work on animations.

  Specifically, this update:

  - Removes timers from the code in `emmy.numerical.ode`. If you want timing
    data you can generate an derivative yourself that performs timing. Including
    this by default introduced a significant performance cost.

  - Renames `emmy.numerical.ode/integration-opts` to `make-integrator*`; it now
    returns only a call to `stream-integrator` instead of the old map of this
    result and timers.

  - In `emmy.expression.compile`:

    - `compile-state-fn*` and `compile-fn*` are now removed, in favor of the
      non-starred versions with option `:cache? false` supplied.

    - `compile-fn` now calls `compile-state-fn` with `params` set to `false` and
      an `:arity` option supplied, vs using its own mostly-duplicated
      implementation.

    - `:flatten?` option removed in favor of the more granular
      `:calling-convention`. This option supports values of `:structure`,
      `:flat` and `:primitive`. See the docstring for details.

    - `:mode` supports `:js`, `:source`, `:clj`, `:native` or `:sci`.

### DIFF
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -292,13 +292,17 @@
 (defn- append-bindings
   "Returns a new let expression with the supplied `bindings`
   wrapping `exp`. If `exp` is already a let expression, the
-  `bindings` will be appended to the existing ones. Each binding
+  `bindings` will be appended (or prepended, if `prepend?`)
+  to the existing ones. Each binding
   is a two-element list [var value]."
-  [exp bindings]
+  [exp bindings & prepend?]
   (let [z (z/seq-zip exp)
         unpaired-bindings (for [kv bindings a kv] a)]
     (z/root (if (= (z/node (z/next z)) `let)
-              (z/edit (z/right (z/next z)) into unpaired-bindings)
+              (z/edit (z/right (z/next z)) (fn [existing-bindings]
+                                             (if prepend?
+                                               (into [] (concat unpaired-bindings existing-bindings))
+                                               (into existing-bindings unpaired-bindings))))
               (z/replace z `(let ~(vec unpaired-bindings) ~(z/node z)))))))
 
 (defn- cse
@@ -333,7 +337,8 @@
       :primitive (let [[ys _ ps] argv]
                    (update code :body append-bindings
                            (concat (local-vars-from-array ys (flatten state-model))
-                                   (local-vars-from-array ps params))))
+                                   (local-vars-from-array ps params))
+                           :prepend))
       code)))
 
 (defn- primitive-body
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -441,24 +441,42 @@
   [{:keys [argv body]}]
   (let [argv    (mapv commafy-arglist argv)
         buffer  (atom [])
+        ;; do-body accepts a zipper location. If the location is a `(doto)` form,
+        ;; it is assumed to be of the form `(doto array (aset index value)...)` and is
+        ;; then unrolled into the corresponding Javascript array mutation statements
+        ;; which are appended to `buffer`. If not, a return statement is generated for
+        ;; the expression at `z`. Nothing is returned.
         do-body (fn [z]
                   (if (and (z/branch? z) (= (z/node (z/next z)) `doto))
                     (let [z (z/next (z/next z))
                           var (z/node z)]
-                      (doseq [[_ ix value] (z/rights z)]
+                      (when-not (symbol? var)
+                        (throw (ex-info "Expecting a symbol (referring to a primitive array)"
+                                        {:unexpected var})))
+                      (doseq [[aset ix value] (z/rights z)]
+                        (when-not (= aset 'aset-double)
+                          (throw (ex-info "Expecting an aset-double statement"
+                                          {:unexpected aset})))
                         (swap! buffer conj (str "  " var "[" ix "] = " (render/->JavaScript value) ";")))
                       (z/next z))
                     (swap! buffer conj
                            (str "  return " (render/->JavaScript (z/node z)) ";"))))
+        ;; do-let accepts a zipper location. If the location is a `(let)` form,
+        ;; the bindings are unrolled into the corresponding sequence of assignment
+        ;; statements which are then appended to `buffer`. If the RHS of a binding
+        ;; pair is of the form `(aget var index)` this will be converted to an
+        ;; array reference; otherwise the usual Javascript renderer is used for the
+        ;; RHS of the assignment. The zipper position immediately following the let
+        ;; form is returned, facilitating this function's use in a parsing chain.
         do-let (fn [z]
-          (if (and (z/branch? z) (= (z/node (z/next z)) `let))
-            (let [z (z/next (z/next z))]
-              (doseq [[var value] (partition 2 (z/node z))]
-                (if (and (seq? value) (= (first value) `aget))
-                  (swap! buffer conj (str "  const " var " = " (nth value 1) "[" (nth value 2) "];"))
-                  (swap! buffer conj (str "  const " var " = " (render/->JavaScript value) ";"))))
-              (z/next z))
-            z))
+                 (if (and (z/branch? z) (= (z/node (z/next z)) `let))
+                   (let [z (z/next (z/next z))]
+                     (doseq [[var value] (partition 2 (z/node z))]
+                       (if (and (seq? value) (= (first value) `aget))
+                         (swap! buffer conj (str "  const " var " = " (nth value 1) "[" (nth value 2) "];"))
+                         (swap! buffer conj (str "  const " var " = " (render/->JavaScript value) ";"))))
+                     (z/next z))
+                   z))
         ]
     (-> body z/seq-zip do-let do-body)
     (conj argv (s/join "\n" @buffer))))
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -489,7 +489,10 @@
   current runtime environment."
   [code]
   #?(:clj (eval (compile->clj code))
-     :cljs (comp js->clj (apply js/Function (compile->js code)))))
+     :cljs (let [g (apply js/Function (compile->js code))]
+             (if (= (:calling-convention code) :primitive)
+               g
+               (comp js->clj g)))))
 
 (defn- compile-sci
   "Returns a Clojure function evaluated using SCI. The returned fn implements
@@ -609,7 +612,7 @@
                                    cache?]
                             :or {mode *mode*
                                  calling-convention :structure
-                                 generic-params? true
+                                 generic-params? (boolean params)
                                  gensym-fn (a/monotonic-symbol-generator 4)
                                  deterministic? false
                                  cache? true}}]
--- a/src/emmy/expression/compile.cljc
+++ b/src/emmy/expression/compile.cljc
@@ -6,6 +6,7 @@
   functions."
   (:require [clojure.string :as s]
             [clojure.walk :as w]
+            [clojure.zip :as z]
             [emmy.expression.analyze :as a]
             [emmy.expression.cse :refer [extract-common-subexpressions]]
             [emmy.expression.render :as render]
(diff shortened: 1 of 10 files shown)


littleredcomputer:
@sritchie once asked "why not just use Clojure as the representation of code and translate everything to JS at the last minute" and this PR is proof that that was a good idea. `render/->JavaScript` still works at the expression level only, but we have a zipper based JS rewriter that uses this lower-level function to write the function body with CSE and primitive state/param extraction and array-based return value. This gets wired up into the ODE solver since we now compile to a function just like what the solvers want and the results are promising.

sritchie:
Amazing, I will review this tomorrow and kick the tires!

src/emmy/expression/compile.cljc
```diff
@@ -111,6 +112,8 @@
                              (- ~'x (~'Math/floor ~'x)))
                      :fn (fn [^double x]
                            (- x (Math/floor x)))}
+   'aset-double #?(:clj {:sym 'clojure.core/aset-double :f aset-double}
```
sritchie, line 115: CRAZILY I think `aset-double` is much slower than `aset`:

```clj
(let [arr ^doubles (double-array 4)
      n 10000000]
  (time (dotimes [_ n]
          (aset-double arr 0 1.0)))
  (time (dotimes [_ n]
          (aset arr 0 1.0))))
"Elapsed time: 571.995333 msecs"
"Elapsed time: 6.592667 msecs"
```

I think because it is coercing its argument each time? I would recommend that we stick with `aset` here...

src/emmy/expression/compile.cljc
```diff
@@ -223,25 +234,147 @@
...
+  This is the typical signature for a state derivative discussed in
+  SICM. It maps a state tuple to its time derivative; structure in,
+  structure out.
+
+  - `:flat => (fn [[t x0 x1 v0 v1] [p0 p1 ...]] ...)`
+
+  ODE solvers prefer to work with flat vectors, and so this calling
+  convention unrolls the state model into an ordered list of parameters.
+
+  - `:primitive => (fn [ys yps ps])`
```
sritchie, line 263: what does yps stand for?

src/emmy/expression/compile.cljc
```diff
@@ -223,25 +234,147 @@
...
+      :primitive (let [[ys _ ps] argv]
+                   (update code :body append-bindings
+                           (concat (local-vars-from-array ys (flatten state-model))
+                                   (local-vars-from-array ps params))
+                           :prepend))
+      code)))
+
+(defn- primitive-body
+  "In the case of primitive calling convention, unrolls the source of the
+  vector-producing statement of the body into a sequence of `aset-double`
```
sritchie, line 346: see `aset-double` note above...

src/emmy/expression/compile.cljc
```diff
@@ -296,61 +429,78 @@
...
-   sym->resolved-form
-   (if (seq bindings)
-     `(fn [~@argv]
-        (let ~(into [] cat bindings)
-          ~body))
-     `(fn [~@argv]
-        ~body))))
+  - `:body`: a function body making use of any symbol in argv above"
+  [{:keys [argv body]}]
+  (w/postwalk-replace sym->resolved-form `(fn [~@argv] ~body)))
```
sritchie, line 440: `sym->resolved-form` smells like the next keyword option, probably for another PR. Some other dictionaries of math that I would use:

- https://github.com/generateme/fastmath/blob/master/src/fastmath/core.clj
- the `emmy.generic` functions, if, for example, I wanted to send a function across the wire and then compile it back to generic
- this is how we might think about compiling, say, matrix-valued functions (that wouldn't work with the primitive calling convention, but it would be great with the other options!)

src/emmy/expression/compile.cljc
```diff
@@ -296,61 +429,78 @@
...
-        code        (s/join
-                     (concat
-                      (for [[var val] bindings]
-                        (str "  const " var " = " (render/->JavaScript val) ";\n"))
-                      ["  return " (render/->JavaScript body) ";"]))]
-    (into [] (concat argv [code]))))
+  its code object argument in a form suitable for application of the Function
+  constructor."
+  [{:keys [argv body]}]
+  (let [argv    (mapv commafy-arglist argv)
```
sritchie, line 447: nice, so this gets all of the bindings up at the top.

src/emmy/expression/compile.cljc
```diff
@@ -296,61 +429,78 @@
...
-
-  - `code`: a map containing `bindings` and `body` keys"
-  [argv code]
-  #?(:clj (eval (compile->clj argv code))
-     :cljs (comp js->clj (apply js/Function (compile->js argv code)))))
+  "Dispatches the `code` object to the compiler corresponding to the
+  current runtime environment."
+  [code]
+  #?(:clj (eval (compile->clj code))
+     :cljs (comp js->clj (apply js/Function (compile->js code)))))
```
sritchie, line 494: do we still need this `js->clj` in every case? Not in the primitive calling convention case, right?

src/emmy/expression/compile.cljc
```diff
@@ -481,53 +669,23 @@
...
-       (do
-         (log/info "compiled function cache hit - arity " n ", mode " mode)
-         cached)
-       (binding [*mode* mode]
-         (let [compiled (compile-fn* f n)]
-           (swap! fn-cache assoc [f n mode] compiled)
-           compiled))))))
+   (compile-fn f n {}))
+  ([f n opts]
+   (let [argv (into [] (repeatedly n #(gensym 'x)))]
```
sritchie, line 688: If we make parameters a keyword option, do you think we could merge `compile-fn` and `compile-state-fn` into a single thing? `compile-state-fn` would be a `compile-fn` call with parameters supplied, otherwise the function's treated as one level deep?

sritchie (changes requested):
This is all looking really, really good... few comments at the edges, but this looks like it's ready already for some performance testing from the animations.

sritchie:
small request after staring at the animation code... could we add a `continuation` return type, that lets the user pass in an `emit` function which should receive all of the arguments that the native mode sets on an output vector?

src/emmy/expression/compile.cljc
littleredcomputer: "y primes" (maybe I should spell it `yprimes`

src/emmy/expression/compile.cljc
littleredcomputer: you're completely right about `aset-double` and I'm removing it everywhere

src/emmy/numerical/ode.cljc
```diff
@@ -186,19 +186,24 @@
...
                            (us/start evaluation-time)
-                           (swap! evaluation-count inc)
-                           (flatten-into-primitive-array out (derivative-fn y))
-                           (us/stop evaluation-time))
+                           (derivative-fn y out primitive-params)
+                           (us/stop evaluation-time)
+                           (swap! evaluation-count inc))
 
         integrator (stream-integrator equations 0 flat-initial-state opts)]
     {:integrator integrator
```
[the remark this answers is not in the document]
littleredcomputer: I think this conclusion is correct.

src/emmy/numerical/ode.cljc
```diff
@@ -173,7 +173,7 @@
 
 
 
-(defn integration-opts
+(defn- integration-opts
```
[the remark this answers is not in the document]
littleredcomputer: I do not know why I changed this: maybe a prompt from the linter but all seems well with `:no-doc`

test/emmy/examples/double_pendulum_test.cljc
```diff
@@ -157,12 +158,13 @@
...
-              (c/compile-state-fn*
+              (c/compile-state-fn
                #(e/Hamiltonian->state-derivative
```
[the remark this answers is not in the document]
littleredcomputer: I like this idea: the whole params thing is a complication and allowing them to vary currently requires departing from the book style of integration. We need a new interface for this. My current thought is that your animation code would appeal to `compile-state-fn` and `stream-integrator` directly, because `evolve` and `state-advancer` do not give you a way to connect the primitive array that react is guarding to the integrand. So my current plan is to say that users of the SICM book interface cannot get the benefit of `generic-params`.

But you may have suggested a way out with the additional keyword parameter, thusly:

"If you say `generic-params?`, that implies compiling, because the only way to do that is to symbolically un-nest the state-derivative. But you must then supply a primitive array which the compiled integrand will close over  *OR* but then I will return to you an allocated primitive array, over which the integrand has closed, that you can mutate" (my preference would be the former, in the hope that that contract is easier to negotiate with react.

src/emmy/expression/compile.cljc
littleredcomputer: I got off on the wrong track with this because `aset` reflection-warns unless the array is type-hinted doubles and the value is type-hinted double. Once that is done there is no reflection and no temptation to use `aset-double`

src/emmy/numerical/ode.cljc
```diff
@@ -186,19 +186,24 @@
...
                              (do (log/warn "Not compiling function for ODE analysis")
-                                 (let [d:dt (apply state-derivative derivative-args)
-                                       array->state #(struct/unflatten % initial-state)]
-                                   (comp d:dt array->state))))
+                                 (let [d:dt (apply state-derivative derivative-args)]
+                                   (fn [ys yps _]
+                                     (flatten-into-primitive-array yps (d:dt (struct/unflatten ys initial-state))))
+                                   )))
         equations        (fn [_ y out]
+                           ;; TODO: should we consider allowing an option to add a dummy
```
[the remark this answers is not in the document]
littleredcomputer: The double-decker state derivative is annoying to deal with in this case since (in the SICM-book-compatible cases) we want to instrument the inner function that is returned from the application to `derivative-args`. So I'm going to take it out for now--as you say the customer can arrange for it themselves

littleredcomputer:
on the matter of `emit`: the integrator needs the values too. Is the animation code watching function evaluations of f? They points at which the function is evaluated are chosen by the integrator for reasons of its own. We do not make any guarantees about the order or granularity of such invocations.

Another way to handle it would be to wrap the compiled function with a dispatcher.

Maybe that is the way to handle instrumentation as well. You could pass an option, perhaps called `finalize` or `adapt`, an "operator" which would be applied to the compiled function before it was given to the integrator. The contract would be that the operator not do something that would "complicate the caching" (or else you can specify `cache?: false` if you really want to do evil).

With that idea, you could instrument and emit or even pass the composition of the two combinators to  `make-integrator*`.

sritchie:
@littleredcomputer I grew deeply confused when I read your comment... why was emit ever working for me when I wired it in??

After a desperate minute I remembered that my `emit` uses were separate from the ODE. `emit` mode was important for visualizing a manifold, or any place where I was passing some code directly to mathbox so that it could graph out a surface. Mathbox handed me pairs of (theta, phi), and I emitted x,y,z coordinates directly. So rather than creating an output array, having the compiled function mutate it and then calling `(emit (aget ...) (aget ...)` or `(apply emit ...)`, passing in `emit` directly was the most efficient thing to do.

littleredcomputer:
Understood. Then let's have both `:calling-convention` (unmodified) and `:return-convention` which can be `:primitive` which gets you `doto out (aset ...` or "something for which `fn?` returns true" which gives you spread application to the values `(k v1 v2 ...)`. 

The ODE code will select `:primitive` for both, or `:structure`, if `:compile` is falsy. You'd select `:primitive` for `:calling-convention` and `emit` for `:return-convention`.

sritchie:
That sounds great, and I think we now have all of my animation use cases covered. Just in time, since it looks like there will be at least one other talk that's using the animation libraries to go over some physics and math.

littleredcomputer:
> Understood. Then let's have both `:calling-convention` (unmodified) and `:return-convention` which can be `:primitive` which gets you `doto out (aset ...` or "something for which `fn?` returns true" which gives you spread application to the values `(k v1 v2 ...)`.
> 
> The ODE code will select `:primitive` for both, or `:structure`, if `:compile` is falsy. You'd select `:primitive` for `:calling-convention` and `emit` for `:return-convention`.

This is on hold because it's not obvious how to paste up a call to an arbitrary function inside our "compiler". One way to do it would just be to `comp` with the result of a primitive computation, and take the hit of `(apply f out)` instead of forming a function call with the arguments spread out.

PR #118, opened 2023-04-05
Update deps, add js? and simplify?

sritchie:
From the CHANGELOG:

- #118:

  - Fixes problems with the `resolve` calls in `emmy.util.def` triggered by
    vanilla (non-shadow) cljs compilation.

  - Updates `Complex.js` and `Fraction.js` dependencies to their latest NPM
    versions.

  - Updates `odex` to the latest release candidate on NPM.

  - The `cljs` side of `emmy.numerical.ode/stream-integrator` gains a `:js?`
    option that defaults to `true`. If `false`, the returned function will
    return native JS arrays vs converting the return value to Clojure.

  - In `emmy.expression.compile`, `compile-fn` and `compile-state-fn` gain a
    `:simplify?` argument that defaults to `true`. If `false`, compilation will
  not attempt to simplify the function body before proceeding.


sritchie:
I will get a test in before merging...

PR #119, opened 2023-04-07
small ODE / compiler fixes

littleredcomputer:
- remove :flat compilation mode (it was a step toward :primitive, but has no use case now)

- add ability to bring your own array to receive integrated points (odex-js supports this already, this brings parity to the JVM implementation)


sritchie:
Looks great, once the merge conflict is resolved go ahead and merge away!

PR #124, opened 2023-04-17
Clerkify

sritchie:


dev/emmy/sci_extensions.cljs
```diff
@@ -0,0 +1,10 @@
+(ns emmy.sci-extensions
```
sritchie, line 1: we don't need this yet, since none of the viewers are actually using Emmy code on the client side.

PR #138, opened 2023-05-11
Prepare for emmy.js, es6 compatibility tweaks

littleredcomputer:


shadow-cljs.edn
```diff
@@ -10,4 +10,4 @@
            :output-to "target/main/advanced.js"
            :modules {:base {:entries [emmy.env]}}
            :compiler-options {:infer-externs :auto
-                              :optimizations :advanced}}}}
+                              :optimizations :advanced}}}}}
```
sritchie, line 13: whoops, extra paren...

src/emmy/abstract/function.cljc
```diff
@@ -157,12 +157,12 @@
   [f]
   (instance? Function f))
 
-(defn- name
-  "Returns the `-name` field of the supplied [[Function]] object. Errors if any
+(defn- f-name
```
sritchie, line 160: If we change the internal name can we keep this the same, or does that mess with what you're trying to do? I have a mild preference for taking advantage of namespaces, so users can write `function/name` vs `function/f-name`... but not if there is some clash with a JS thing here that we want to avoid.

src/emmy/calculus/derivative.cljc
```diff
@@ -200,25 +200,34 @@
...
+  (perturbed? [_] false)
   (replace-tag [f old new] (replace-tag-fn f old new))
   (extract-tangent [f tag] (extract-tangent-fn f tag))
 
   #?@(:cljs
-      [MetaFn
+      [;; In Clojurescript, we arrange for metadata to live directly on
+       ;; function objects by setting a special property and implementing
+       ;; IMeta and IWithMeta: see [[emmy.value]].
+       function
```
sritchie, line 218: can we still implement `perturbed?` here?

src/emmy/function.cljc
```diff
@@ -12,7 +12,8 @@
             [clojure.core.match :refer [match]]
             [emmy.generic :as g]
             [emmy.util :as u]
-            [emmy.value :as v])
+            [emmy.value :as v]
+            [taoensso.timbre :as log])
```
sritchie, line 16: Remove this if it's not needed

src/emmy/function.cljc
```diff
@@ -73,7 +87,7 @@
   [& fns]
   (let [a (arity (or (last fns)
                      identity))]
-    (with-meta (apply comp fns) {:arity a})))
```
sritchie, line 76: way better!

src/emmy/function.cljc
```diff
@@ -400,8 +408,7 @@
                                                   (cond (= m n) [:exactly m]
                                                         (< m n) [:between m n]
                                                         :else (fail)))
-              [[:between m n] [:exactly k]] (if (and (<= m k)
-                                                     (<= k n))
+              [[:between m n] [:exactly k]] (if (<= m k n)
```
sritchie, line 411: better!!

src/emmy/function.cljc
```diff
@@ -465,12 +472,14 @@
...
-                    (op (apply f1 args)
-                        (apply g1 args)))
-                  (with-meta {:arity arity}))))]
-    (with-meta h {:arity [:exactly 2]})))
+                  arity   (joint-arity [f-arity g-arity])]
+              (case arity
+                [:exactly 1] (fn [a] (op (f1 a) (g1 a)))
+                [:exactly 2] (fn [a b] (op (f1 a b) (g1 a b)))
+                (do
+                  (log/warn "creating a function with metadata: these are tricky to exploit from JavaScript")
```
sritchie, line 480: ah, sorry, we do use log here!

src/emmy/structure.cljc
```diff
@@ -19,7 +19,9 @@
                             AFn IFn
                             IPersistentVector IReduce IKVReduce
                             IObj
-                            Indexed Reversible Sequential))))
+                            Indexed Reversible Sequential)))
+  #?(:cljs
```
sritchie, line 23: Do double require statements work? The pattern I like is sticking a line in the big require block above like

```
#?(:cljs [goog.object :as gobject])
```

src/emmy/structure.cljc
```diff
@@ -46,6 +48,7 @@
 (derive ::up ::structure)
 (derive ::down ::structure)
 (derive #?(:clj IPersistentVector :cljs PersistentVector) ::up)
+#?(:cljs (derive js/Array ::up))
```
sritchie, line 51: good catch...

src/emmy/structure.cljc
```diff
@@ -241,56 +249,91 @@
...
+                (make orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q) v) m))
        (-invoke [_ a b c d e f g h i j k l m-arg n o p q r]
-                (Structure. orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r) v) m))
+                (make orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r) v) m))
        (-invoke [_ a b c d e f g h i j k l m-arg n o p q r s]
-                (Structure. orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r s) v) m))
+                (make orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r s) v) m))
        (-invoke [_ a b c d e f g h i j k l m-arg n o p q r s t]
-                (Structure. orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r s t) v) m))
+                (make orientation (mapv #(% a b c d e f g h i j k l m-arg n o p q r s t) v) m))
```
sritchie, line 292: I really want a macro that will auto-expand out these IFn patterns in our deftypes! We'll get one at some point

src/emmy/structure.cljc
```diff
@@ -489,9 +532,14 @@
...
   "Generate a structure with the supplied orientation, given some sequence `xs`"
-  [orientation xs]
-  (let [xs (if (vector? xs) xs (into [] xs))]
-    (->Structure orientation xs nil)))
+  ([orientation xs]
+   (make orientation xs nil))
+  ([orientation xs meta]
+   (let [xs (if (vector? xs) xs (into [] xs))]
+     (->Structure orientation xs meta)
+     #_(-> (->Structure orientation xs meta)
```
sritchie, line 540: can we remove this if it's not used? Or do we need this?

sritchie (commented):
This is so good... I have a few small changes I want to get in to the next release, so I'll prepare those and we'll pull the trigger after both sets are in.

src/emmy/abstract/function.cljc
littleredcomputer: That's a great idea, thank you for pointing it out. The internal property name is an implementation detail, thank you for reminding me of that.

src/emmy/function.cljc
littleredcomputer: I was using it but have refactored things now to avoid it. This PR has been through a few layers that I'm sanding out. Initially I tried to dodge the fn-with-meta problem by making functions with "manifest arity" in this case and letting the arity cache deal with it in the future, but now we hack the arity into the function objects at creation time which is easier to reason about IMO

src/emmy/function.cljc
littleredcomputer: will retire

src/emmy/structure.cljc
littleredcomputer: obsolete

littleredcomputer: ✅

src/emmy/structure.cljc
littleredcomputer: I toyed with the idea of forcing all structure creation through this point. There is still a debate to be had about that. Making an object a Function that wasn't born that way requires the use of `Object.setPrototypeOf` which is [strongly discouraged](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) at MDN since it frustrates certain optimizations. 

I found that Markus' tests could be done without medicine this strong, by moving these calls to the TypeScript tuple constructors. The possibility remains (and indeed there is one such case among those tests) in which an up-tuple is generated by interior code below the TypeScript layer but pops back out in user land without this decoration. This is rare though.

The decision of whether it is better to treat all structs the same even at the risk of annoying V8 or not is one we might use a spreadsheet for. One intermediate example is to have a *dynamic-var* that controls this behavior, and have the `emmy.js` library opt into this feature by setting that var true. 

I'm trying to protect the performance/principle of what we have (though I don't know if this is any slower). The argument on the other side is "Emmy is a dialect of the host language, and the transparency of its use is going to require things like this; brace yourself."

src/emmy/calculus/derivative.cljc
littleredcomputer: yes (not sure how I overlooked that!)

src/emmy/calculus/derivative.cljc
```diff
@@ -200,25 +200,34 @@
 ;; preserved through tag replacement and extraction.
 
 (extend-protocol d/IPerturbed
-  #?(:clj Fn :cljs function)
-  (perturbed? [#?(:cljs _ :clj f)]
-    #?(:clj (:perturbed? (meta f) false)
-       :cljs false))
+
+  ;; In Clojure, metadata can live directly on function objects.
+  Fn
```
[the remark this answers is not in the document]
littleredcomputer: ✅

src/emmy/structure.cljc
littleredcomputer: this wasn't necessary after all so it's gone anyway.

littleredcomputer:
I rebased the whole change on top of the fixes to the matrix-by-rows reporting.

src/emmy/value.cljc
```diff
@@ -438,6 +439,55 @@
...
+  ;; cloning the function as the `IWithMeta` contract would require,
+  ;; but this implementation should be safe enough for our purposes.
+  ;;
+  ;; By using a symbol instead of a string as the key for the metadata,
+  ;; the metadata annotation will not be visible among the object's other
+  ;; normal properties.
+   (let [metadata-symbol (. js/Symbol for "Symbol.__emmy_meta__")]
+     (extend-type function
+       IMeta
+       (-meta [f] (or (gobject/get f metadata-symbol) {})))
```
sritchie, line 462: I think cljs code prefers "nil" as the no-metadata-attached case... what do you think?

src/emmy/structure.cljc
```diff
@@ -291,6 +296,41 @@
                 (.-v s))
          (print-method w))))
 
+#?(:cljs
+   (defn make-es6-indexable
```
sritchie, line 300: This would be a good candidate for `^:no-doc` metadata so it doesn't appear in the public cljdoc api docs

src/emmy/value.cljc
```diff
@@ -438,6 +439,55 @@
...
+
+    ;; We would like to support IWithMeta, but that interface's contract
+    ;; requires the production of a _new_ object. Instead we provide a function
+    ;; to mutate a native object's metadata in place.
+     (defn set-js-meta!
+       "Mutates the native JS object `o` to have the given metadata. The
+       previous metadata, if any, is discarded. `o` is returned."
+       [o m]
+       (unchecked-set o metadata-symbol m)
```
sritchie, line 471: Could do

```
(doto o
  (unchecked-set metadata-symbol m))
```

but I like yours, `doto` is cleaner with multiple mutations but with 1 this feels clearer

src/emmy/value.cljc
```diff
@@ -438,6 +439,55 @@
...
+       o))
+   )
+
+#?(:cljs
+   (let [property-name-replacements {"name" "-name"}]
```
sritchie, line 476: Speaking from total inexperience...

1. are dashes allowed in JS property names? I know they work for `obj["-name"]` but `obj.-name` is illegal, right? I hit this when realizing that cljs munges `-` to `_`, so if you have `a-x` and `a_x` variables one will shadow the other.
2. I'm not clear on why we need the replacement, so I'll ask if we still need this after the f-name replacement? was that why it was here?

sritchie (approved):
Just a few more comments, looking great! Once you're satisfied, merge away.

src/emmy/value.cljc
littleredcomputer: I have retired this, my only motive was not to drop a property on the floor that might be useful elsewhere, but that doesn't fix an actual problem we have, so dropped it

src/emmy/value.cljc
littleredcomputer: I agree. I like doto iff n > 1

src/emmy/structure.cljc
littleredcomputer: ✅

PR #139, opened 2023-05-11
Fix unit tests, concretize protocol impls

littleredcomputer:
- #139:

  - Fixes matrix and simplification-related test failures introduced in the last
    couple of PRs.

  - Removes the custom `->Quaternion` constructor in `emmy.quaternion`; this was
    causing build warnings in `cljs`. Any place you might have used
    `->Quaternion`, please use `emmy.quaternion/make`.

  - Backs off the various protocol extensions in `emmy.collection` from
    `IPersistentMap` and `IPersistentSet` to the concrete types shipped with
    Clojure. This prevents these implementations from attaching, on the JVM
    side, to new `defrecord` instances or other types that extend those
    protocols.


README.md
```diff
@@ -288,19 +290,20 @@ Distributed under the [GPL v3](LICENSE) license. See [LICENSE](LICENSE).
 [cljdoc]: https://cljdoc.org/badge/org.mentat/emmy
 [clojars-url]: https://clojars.org/org.mentat/emmy
 [clojars]: https://img.shields.io/clojars/v/org.mentat/emmy.svg
-[codecov]: https://img.shields.io/codecov/c/github/mentat-collective/emmy/main.svg?maxAge=3600
```
sritchie, line 291: these links didn't change, I just sorted them.

src/emmy/collection.cljc
```diff
@@ -96,14 +100,9 @@
 ;; structure, or a dataframe-like structure with named fields instead of
 ;; positional fields. Nothing like this exists yet!
 
-#?(:clj
```
sritchie, line 99: @littleredcomputer , I want your thoughts here.

I took advantage initially of the ability on the JVM side to extend protocols to java interfaces. That caused trouble when I realized that defrecord instances were picking up all of my implementations...

So here I am backing out the extension to interface in favor of extending these protocols by default only to the concrete map and set implementations.

I've left the vector and seq implementations generic for no good reason other than that I don't know if I've got all of the various seq types listed on the JVM. Also there is no mechanism like defrecord for building a vector-like thing. But still, if we make a deftype that extends IPersistentVector it would accidentally gain all of these implementations.

Is there any reason I should NOT also concretize seq and vector to be safe?

src/emmy/env.cljc
```diff
@@ -93,7 +93,7 @@
 
   TODO add `:rename`, `:exclude` support."
   []
-  `(require '~['emmy.envxmy
+  `(require '~['emmy.env
```
sritchie, line 96: whoops!

src/emmy/quaternion.cljc
```diff
@@ -404,17 +404,6 @@
        (-invoke [this a b c d e f g h i j k l m n o p q r s t rest]
                 (evaluate this (into [a b c d e f g h i j k l m n o p q r s t] rest)))]))
 
-#_{:clj-kondo/ignore [:redefined-var]}
-(do (ns-unmap 'emmy.quaternion '->Quaternion)
```
sritchie, line 408: this was causing a build error in emmy-viewers. I thought I was being clever by allowing metadata to be smuggled in. But using `make` everywhere that we want to avoid passing in the metadata argument is the proper approach.

test/emmy/collection_test.cljc
```diff
@@ -83,7 +83,13 @@
       (is (every? v/zero? (v/zero-like (list 1 2 3)))
           "works with lists"))))
 
+(defrecord MyRecord [])
+
 (deftest map-tests
+  (testing "don't try and simplify records"
```
sritchie, line 89: I had initially found this issue by trying to print `emmy.pattern.match/failure` in Cider, which I route through `expression->stream`.

test/emmy/generic_test.cljc
```diff
@@ -146,15 +146,13 @@
...
-;; TODO fix this in cljs.
-#?(:clj
-   (deftest simplify-tests
-     (checking "g/simplify" 100 [k gen/keyword]
-               (is (= k (g/simplify k)) "keywords simplify to themselves."))
-
-     (is (= [:div "cake"]
-            (g/simplify [:div "cake"]))
-         "Keywords simplify to themselves inside of nested structures")))
+(deftest simplify-tests
```
sritchie, line 149: fixed!

sritchie (commented):
@littleredcomputer , some notes for you...

src/emmy/collection.cljc
littleredcomputer: I agree with your decision to concretize seq and vector. Rationale: it should be safe for J. Random Thing to implement these interfaces without silently also becoming part of our system. The user can opt another class into the system without requiring anything from us. If someone feels we have overlooked something they can let us know.

littleredcomputer (commented):
your changes lgtm

src/emmy/generic.cljc
```diff
@@ -563,8 +563,8 @@ Equivalent to `(invert (sin a))`."
 
 (defgeneric atan [1 2]
   "Computes the inverse tangent of the supplied argument `a`. Given two
-  arguments `a` and `a`, returns the inverse tangent of the angle formed by the
-  point `(a, b)` in a 2-dimensional euclidean plane.
+  arguments `a` and `b`, returns the inverse tangent of the angle formed by the
+  point `(b, a)` in a 2-dimensional euclidean plane.
```
sritchie, line 567: whoops, `a` and `b` are poor names here but at least this matches reality now. @littleredcomputer.

PR #143, opened 2023-07-03
Littleredcomputer/cse

littleredcomputer:


src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+;; but should be perfectly transparent to the code generators of the host
+;; languages.)
+
+(defn- dissociate
+  "If `x` is an expression like `(f x1 x2 x3...)`, where `(f? f)` is true,
+  returns `(f ... (f x1 x2) x3 ...)`, i.e., we replace applications with
+  arity > 2 with left-associative nested applications of arity 2. Otherwise
+  `x` is returned."
+  [f? x]
+  (letfn [(go [x]
```
sritchie, line 61: Couple of ideas, not sure if they are helpful at all (your call either way!)

```clojure
(defn- dissociate [f? x]
  (letfn [(go [x]
              (cond (not (seq? x)) x
                    
                    ;; I am irrationally scared of laziness in cases like this, after being bitten in the simplifier
                    ;; implementation by a lazy `map`. Maybe it doesn't matter here, but you could `doall`...
                    (or (not (f? (first x)))
                        (< (count x) 4))
                    (doall (map go x))
                    
                    :else
                    ;; transduce would work here too and save an intermediate sequence... wdyt?
                    (let [[f x & xs] x]
                      (transduce (partial list f)
                                 (map go)
                                 x
                                 xs))))]
    (go x)))
```

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+    (go x)))
+
+;; This is the heart of the F90 algorithm. We do a post-order traversal
+;; of the tree, building an association between equivalence classes of
+;; subtree to small integers. We can then represent the maximally factored
+;; DAG as a simple sequence of expressions, where integers in argument
+;; position represent back-references to the indexed computation emitted
+;; previously (that is, in the nth expression, indices in the range [0..n-1]
+;; may appear). Thus the computation may be efficiently performed by
+;; evaluation the subexpressions in the order reported.
```
sritchie, line 77: "by evaluating"

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+              (or (@table x)
+                  (let [i (swap! counter inc)]
+                    (swap! table assoc x i)
+                    i)))
+            (uid [x]
+                (cond
+                  (symbol? x) x
+                  (number? x) (install x)
+                  (seq? x) (install (into [(first x)] (map uid (next x))))
+                  :else (throw (u/exception (str "cse? " x)))))]
```
sritchie, line 105: I think `u/exception` already throws, no need for the `throw` wrapper

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+  by new expressions; in this way, multiple expressions in the same scope
+  may be compacted.
+
+  Passing the argument `:table` will result in a sequence
+  of simple forms with a function symbol at the head followed by arguments.
+  The arguments are either symbols drawn from the original expression or
+  integers which represent the the index of the value of a computation
+  performed earlier in the sequence."
+  []
+  (let [counter (atom -1)
```
sritchie, line 93: could use a `volatile!` here, `(volatile! -1)`, and then `vswap!` below...

From https://github.com/clojure/clojure/blob/97c1210713cbbe0bacbb59004a91e25515d74ab6/changes.md#21-transducers :

> volatiles - there are a new set of functions (volatile!, vswap!, vreset!, volatile?) to create and use volatile "boxes" to hold state in stateful transducers. Volatiles are faster than atoms but give up atomicity guarantees so should only be used with thread isolation.

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+        table   (atom {})]
```
sritchie, line 94: A volatile would be faster here too, see the distinct transducer for a similar spot that uses this: https://github.com/clojure/clojure/blob/master/src/clj/clojure/core.clj#L5075

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+    (letfn [(install [x]
+              (or (@table x)
+                  (let [i (swap! counter inc)]
+                    (swap! table assoc x i)
+                    i)))
+            (uid [x]
+                (cond
+                  (symbol? x) x
+                  (number? x) (install x)
+                  (seq? x) (install (into [(first x)] (map uid (next x))))
```
sritchie, line 104: Nitpick, prefer the transducer form of 

```clojure
(seq? x) (install (into [(first x)] (map uid) (next x)))
```
just one paren moved!

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+
+(defn- uid-assigner
+  "Produces a function which takes an S-expression and produces the
+  compacted DAG representation of the tree, as described in [Flajolet 90].
+  Returns the ID of the root of the tree. The assigner generated by this
+  function can be used multiple times, and earlier compact forms will be reused
+  by new expressions; in this way, multiple expressions in the same scope
+  may be compacted.
+
+  Passing the argument `:table` will result in a sequence
```
sritchie, line 87: if you don't want the magic keyword, you could trigger this behavior with a no-arg call?

src/emmy/expression/cse.cljc
```diff
@@ -7,152 +7,154 @@
...
+                    i)))
+            (uid [x]
+                (cond
+                  (symbol? x) x
+                  (number? x) (install x)
+                  (seq? x) (install (into [(first x)] (map uid (next x))))
+                  :else (throw (u/exception (str "cse? " x)))))]
+      (fn [expr]
+        (if (= expr :table)
+          (->> @table (sort-by second) (map first))
```
sritchie, line 108: this is another case where I will recommend a `doall`, given that this is querying a stateful data structure and we probably want the query to trigger at the time of the call.

sritchie (commented):
Looks really good, and I have been playing with the results over at emmy-viewers. I want to stare a bit more at the algorithm tomorrow but here are my first responses...

src/emmy/expression/compile.cljc
```diff
@@ -341,37 +341,24 @@
...
+  (letfn [(children? [x]
+            (and (sequential? x) ('#{up down vector} (first x))))
+          (aset-form [index value]
+            `(aset ~index ~value))]
+    (if (= calling-convention :primitive)
+      (update code :body
+              (fn [body]
+                (if (children? body)
+                  (let [array-symbol (nth argv 1)
+                        values (filter (complement children?) (tree-seq children? rest body))]
```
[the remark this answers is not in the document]
littleredcomputer: done.

test/emmy/expression/compile_test.cljc
```diff
@@ -130,29 +154,53 @@
 
 
           (doseq [[mode expected-source]
-                  {:clj `(fn ~'[[y1 [y2 [y3 [y4]]]]]
-                           (vector (+ (* 3.1 ~'y1) 1.55)))
-                   :js ["[y1, [y2, [y3, [y4]]]]" "  return [3.1 * y1 + 1.55];"]}]
+                  {:clj '(fn [[y1 [y2 [y3 [y4]]]]]
+                           (let [_5 3.1
+                                              _6 (* _5 y1)
```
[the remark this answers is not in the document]
littleredcomputer: ✅

src/emmy/expression/cse.cljc
littleredcomputer: But the value of `@table` is immutable?

src/emmy/expression/cse.cljc
littleredcomputer: ✅

src/emmy/expression/cse.cljc
sritchie: oh, yeah, good call! Told you I had paranoia.

src/emmy/expression/cse.cljc
littleredcomputer: thank you. I have to get better at using transducers.

src/emmy/expression/cse.cljc
littleredcomputer: We have the thread isolation that is called for, but my own preference would be to reserve volatile for situations where the mutable thing is created and disposed of within the function application itself even though this is unnecessarily conservative as you point out.

src/emmy/expression/cse.cljc
littleredcomputer: see below

src/emmy/expression/cse.cljc
littleredcomputer: ✅

src/emmy/expression/cse.cljc
littleredcomputer: ✅

src/emmy/expression/cse.cljc
littleredcomputer: I think it is reasonable for this function to be lazy, which would be useful if this were one in a sequence of tree-transformations, and that might happen as the compiler gets more powerful. Looking at where the laziness is capped in this case, it's an `into []` enforced by its caller.

`transduce` hint gratefully accepted!

littleredcomputer:
```clojure
(transduce (map str) (partial list 'f) 'x [2 3 4])
(f (f (f (f x "2") "3") "4"))
```
Annoyingly an arity-1 call to f is used to finalize the reduction (grr). I think reduce is more transparent here

src/emmy/expression/compile.cljc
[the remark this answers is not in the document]
littleredcomputer: ✅

src/emmy/expression/cse.cljc
littleredcomputer: @sritchie points out that `completing` would fix that problem, but I've stuck with reduce on the KISS principle

PR #144, opened 2023-07-25
sci-macros

mhuebert:
- Wrap macros that are intended to be exposed to sci, in a new `sci-macro` macro. During macro expansion, when `(:ns &env)` is true, a function is returned `^:macro` metadata, as well as `&env` and `&form` args added to each arity.
- use sci.core/copy-ns to copy namespaces, using the above wrapped macros to avoid sci-specific macro implementations

with `bb test:clj`, `emmy.sci-test` tests are passing. Some others are failing but this seems to be the case in `main` as well.

still working on cljs tests to pass - problems with `careful-def` remain


src/emmy/env.cljc
```diff
@@ -97,7 +97,7 @@
                :refer
                (into [] (keys (ns-publics 'emmy.env)))]))
 
-(defmacro literal-function
+(u/sci-macro literal-function
```
sritchie, line 100: can the new `import-def` handle macros? If so I can delete these and move them down into the `import-vars` block below... so good!

src/emmy/util/def.cljc
```diff
@@ -220,10 +240,15 @@
   (In ClojureScript, only forms like `(def ~sym ~form)` are emitted, since the
   compiler does not currently error in case 2 and already handles emitting the
   warning for us.)"
-  [#?(:clj ns :cljs _)]
+  [env ns]
   #?(:cljs
```
sritchie, line 244: phew, this deserves to be pushed back into potemkin or live somewhere that folks can use it (and invoke grumbling on clojurians slack of course)

sritchie (approved):

src/emmy/calculus/coordinate.cljc
```diff
@@ -156,22 +156,21 @@
   - `dx` and `dy` bind to 1-forms for each coordinate."
   [coordinate-prototype coordinate-system]
   (let [sys-name           (symbol (name coordinate-system))
+        value-sym          (symbol (str sys-name "-values"))
```
sritchie, line 159: I suspect we should be using a `gensym` here...

sritchie:
Closed in favor of #145 .

PR #149, opened 2023-10-10
Retire value protocol in favor of multimethods

littleredcomputer:
`Value` is now renamed `IKind` since `kind` is now the only method that inhabits the protocol.

The other methods have been made into Emmy generics with `defmulti`.


CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
...
+      protocol instead of defining a method that throws. The exception thrown
+      in such cases therefore changes from UnsupportedOperationException to
+      IllegalArgumentException.
+
+    - For this reason we have avoided defining `:default` handlers for
+      generic MultiFns, despite the fact that defaults are certainly
+      convenient in some cases (like `false` for `exact?` and having
+      the default case for `freeze` pass through).
+
+    - The default exponentiation routine built on generic multiplication
```
sritchie, line 41: Good catch!

src/emmy/abstract/function.cljc
```diff
@@ -230,9 +220,9 @@
           litfns)))
 
 (u/sci-macro with-literal-functions [litfns & body]
```
sritchie, line 222: We definitely want this indentation to match `defmacro`... can you try adding

```
{:clj-kondo/lint-as 'clojure.core/defn}
```

to `emmy.util/sci-macro` to see if vscode picks it up? 

Of course

```
{:clj-kondo/lint-as 'clojure.core/defmacro}
```

makes more sense but I think `defn` might be the only supported option...

src/emmy/abstract/function.cljc
```diff
@@ -311,3 +301,11 @@
               (f/arity f)
               (domain-types f)
               (range-type f)))
+
+(defmethod g/zero-like [::function] [^Function a] (fn [& _] (g/zero-like (.-range a))))
```
sritchie, line 305: does the range always have an instance of the output type? I thought we could have things like `'(UP Real Real)`.

I have a PR bringing in the Sussman-style here: https://github.com/mentat-collective/emmy/pull/113

I wonder if this implementation will work there, or what you think...

src/emmy/abstract/function.cljc
```diff
@@ -311,3 +301,11 @@
...
+(defmethod g/one-like [::function] [^Function a] (fn [& _] (g/one-like (.-range a))))
+(defmethod g/identity-like [::function] [^Function a]
```
sritchie, line 307: is it worth making this identity a top-level var and then returning that here/

src/emmy/abstract/function.cljc
```diff
@@ -311,3 +301,11 @@
...
+  (let [meta {:arity (.-arity a) :from :identity-like}]
+    (with-meta identity meta)))
+(defmethod g/exact? [::function] [a] (f/compose g/exact? a))
```
sritchie, line 310: hmm, won't this always return symbols, which will register as false? I guess that makes sense!

src/emmy/calculus/map.cljc
```diff
@@ -11,9 +11,9 @@
             [emmy.calculus.manifold :as m]
             [emmy.calculus.vector-field :as vf]
             [emmy.function :as f]
+            [emmy.generic :as g]
```
sritchie, line 14: so nice to see this unified

src/emmy/collection.cljc
```diff
@@ -128,9 +122,9 @@
 
 (defn- combine [f m1 m2 l-default]
   (letfn [(merge-entry [m e]
-			      (let [k (key e)
```
sritchie, line 131: haha, these were left over from a copy-paste from scheme before I converted, clearly... I'm glad you caught one!

src/emmy/collection.cljc
```diff
@@ -235,33 +220,23 @@
   (cs/union a b))
 
 (doseq [klass [PersistentHashSet PersistentTreeSet]]
+  (defmethod g/zero? [klass] [s] (empty? s))
```
sritchie, line 223: I'm counting on our tests to catch my tired eyes as I read these implementations... this is heroic work!

src/emmy/differential.cljc
```diff
@@ -13,6 +13,7 @@
   (:refer-clojure :exclude [compare])
   (:require [clojure.core :as core]
             [clojure.string :refer [join]]
+            [emmy.function]  ;; for the side effect of making kind: MultiFn -> ::v/function
```
sritchie, line 16: ::g/function?

src/emmy/expression/analyze.cljc
```diff
@@ -19,6 +19,7 @@
   an expression with respect to an analyzer is therefore effected by a
   round-trip to and from the canonical form."
   (:require [emmy.expression :as x]
+            [emmy.generic :as g]
             [emmy.numsymb :as sym]
             [emmy.util :as u]
             [emmy.value :as v]))
```
sritchie, line 25: can this import go away or no?

src/emmy/ratio.cljc
```diff
@@ -12,7 +12,7 @@
   (:refer-clojure :exclude [ratio? numerator denominator rationalize])
   (:require #?(:clj [clojure.core :as core])
             #?(:clj [clojure.edn] :cljs [cljs.reader])
-            #?(:cljs ["fraction.js/bigfraction.js$default" :as Fraction])
+            #?(:cljs ["fraction.js/bigfraction.js" :as Fraction])
```
sritchie, line 15: interesting, why do JS people keep messing with their builds...

src/emmy/simplify.cljc
```diff
@@ -74,7 +74,7 @@
         expr
         (let [canonicalized-expr (canonicalize new-expr)]
           (cond (= canonicalized-expr expr) expr
-                (v/zero?
+                (g/numeric-zero?
```
sritchie, line 77: a change!!

sritchie (commented):
Still reviewing but here is my first batch of comments, @littleredcomputer !

src/emmy/value.cljc
```diff
@@ -331,7 +246,7 @@
 
      IPrintWithWriter
      (-pr-writer [x writer _]
-       (let [rep (if (<= x (.-MAX_SAFE_INTEGER js/Number))
+       (let [rep (if (< (if (< x 0) (- x) x) (.-MAX_SAFE_INTEGER js/Number))
```
sritchie, line 249: good catch! I might recommend a test for this if it didn't make it in yet, which it might have, apologies for my code nav on this PR

sritchie (approved):

sritchie:
@littleredcomputer looking good pending the few comments that I left... thank you for doing this!!

CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
...
+      would, in the case of raising to the zero power, return the one-like
+      of the exponent; now it returns the one-like of the base, which is
+      correct.
+
+    - The generic `negative?` required Numbers instead of Comparables;
```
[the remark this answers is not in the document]
littleredcomputer: Yup! I was inspired by similar examples you had designed.

package.json
```diff
@@ -25,7 +27,7 @@
     "@nextjournal/lezer-clojure": "1.0.0",
     "complex.js": "^2.1.1",
     "d3-require": "1.3.0",
-    "fraction.js": "^4.2.0",
+    "fraction.js": "^4.3.6",
```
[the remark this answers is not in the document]
littleredcomputer: I'll talk more about this when we get to the patch...

patches/fraction.js+4.3.7.patch
```diff
@@ -0,0 +1,22 @@
...
++  "type": "commonjs",
+   "main": "fraction.cjs",
+   "exports": {
+     ".": {
+       "import": "./fraction.js",
+       "require": "./fraction.cjs",
+       "types": "./fraction.d.ts"
+-    }
++    },
++    "./bigfraction.js": "./bigfraction.js"
```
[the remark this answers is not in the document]
littleredcomputer: So: I changed nothing about fraction.js that seemed material, but CI complained of being unable to import when I pushed to GH. This was frustrating, but it did seem that the complaint on the CI side was justified. Maybe there is a difference between the shadow-cljs version on my machine vs. what CI is running, but I thought that it would be safer to fix this problem "once and for all" (see below)

src/deps.cljs
```diff
@@ -1,4 +1,4 @@
 {:npm-deps
  {"complex.js" "^2.1.1"
-  "fraction.js" "^4.2.0"
+  "fraction.js" "^4.3.6"
```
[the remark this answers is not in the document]
littleredcomputer: That's a good comment, and the developers of Fraction.js do suggest just "vendoring" the bigfraction.js if that's the API you want. I'm not sure quite where to put it.

On the other hand, the patches get swept in by the post install hook in package.json, so this feels like a principled approach...but if it does frustrate consumers of Emmy then it has to go.

src/emmy/collection.cljc
littleredcomputer: "Refactor in the presence of running tests" has never felt so relevant!!

src/emmy/expression/analyze.cljc
littleredcomputer: there's a ref. to `g/freeze` and `v/integral?`

Some of the utility level stuff in v/ could move to g/, and then the import of v would occur less often. Willing to consider that

src/emmy/ratio.cljc
littleredcomputer: I did notice that (1) in the current version, there's an exports block in fraction.js's package.json, (2) bigfraction.js  is not listed there, and (3) I still don't know why this was ever a problem on GitHub, where the older version had no exports clause and I guess was relying on a permissive interpretation.

I wanted to get "everyone onto the same page" and so experimented with adding and removing `$default`. I don't even know what it's really for, I was just trying to shake things into place. 

I'm quite open to the idea of just stashing bigfraction.js somewhere in the tree and forgetting about it

src/emmy/simplify.cljc
littleredcomputer: This is the one where what's being tested for zero is a lazy seq returned from simplify that got me wondering about this... I wanted to avoid the idea that a realized empty list could spoof this. I can put it back if you prefer

littleredcomputer: BTW, I think what you had originally written is perfectly safe, in hindsight

src/emmy/value.cljc
littleredcomputer: Something definitely breaks w/o this :) 

What has happened, I think, is that there were some "overlapping" cases in the Value regime that have been clarified by the transition to defmulti, which focused my attention on the (very few) bugs that I found

littleredcomputer:
> @littleredcomputer looking good pending the few comments that I left... thank you for doing this!!

I'm really happy to do this

src/deps.cljs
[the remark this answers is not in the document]
littleredcomputer: I experimented using a local "vendored" version of bigfraction.js: doing that is easy enough but it needs a package.json to go with it. Fine, but that seems morally equivalent to the patch-package technique I've used. Since CLJS compilation can't happen without `nom i`, and that will take care of applying the patch, are you still worried?

src/emmy/differential.cljc
littleredcomputer: I think `::v/function` is the correct spelling here

src/emmy/value.cljc
```diff
@@ -162,98 +142,33 @@
...
-  (zero? [_] false)
-  (one? [_] false)
-  (identity? [_] false)
-  (zero-like [v] (u/unsupported (str "zero-like: " v)))
-  (one-like [v] (u/unsupported (str "one-like: " v)))
-  (identity-like [v] (u/unsupported (str "identity-like: " v)))
-  (freeze [v] (:name (meta v)))
-  (exact? [_] false)
-  (kind [v] (type v))
+;;  Var
```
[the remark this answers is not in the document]
littleredcomputer: Currently we just have `kind` for nil (returning nil). Since nil no longer gets any of the other formerly-Value methods, I thought, should it have a `kind`?

If nil has no kind, i.e., it's not an Emmy value at all, a few things break. We lose this:

```
  (defmethod g/partial-derivative [t nil] [f _]
    (multivariate f [])))
```

Although I think that could be fixed by having the `(kind nil)` be `::v/seqtype`. There are a few other cases of error that we could fix.

But the experiment of removing `IKind` from `nil` has kind of convinced me that we want the `kind` function to be a total function, safe to call on any Clojure(Script) value, so that a defmethod could in principle be written for any object.

So I am comfortable with the status quo. Not having the other Value methods like `zero?` defends against unfortunate nil-puns.

src/deps.cljs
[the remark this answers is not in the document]
sritchie: My concern is about downstream libraries like Emmy-viewers - will all dependents now have to configure this patch as well? I would be fine pinning to an older version of Fraction as well to avoid this!

littleredcomputer:
Finally have enough coverage. Now on to fixing the fraction.js thing and we're done.

package.json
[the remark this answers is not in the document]
littleredcomputer: pinning version to 4.2.1 seems to work. Pulling the patch back out of the build. A battle for another day (or we can use your template and just supply the implementation)

patches/fraction.js+4.3.7.patch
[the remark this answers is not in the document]
littleredcomputer: obsolete

PR #151, opened 2023-10-20
Implement Complex numbers natively in cljs

sritchie:
- #151

  - By porting `complex.js` to Clojure, we can remove the dependency on
    this library on the JavaScript side as well as the Apache Commons
    complex implementation on the JVM side.

    The JavaScript implementation is followed fairly closely and done with
    generic Emmy arithmetic at the component level (except when that is
    clearly unnecessary). The `(complex)` constructor has been made
    equivalent to the reader parser.

    The former implementation made a special case of i^r, raising the
    complex unit to a real power, but it only worked for integral r, and
    threw an exception in other cases; this special case is removed.

### DIFF
--- a/src/emmy/complex.cljc
+++ b/src/emmy/complex.cljc
@@ -75,29 +75,14 @@
 (defn round
   "Generates a [Gaussian integer](https://en.wikipedia.org/wiki/Gaussian_integer)
   from the complex number `z` by rounding the real and imaginary components of
-  `z` to their nearest integral values."
+  `z` to their nearest integral values. (Note: the use of cast-to-double is
+   unfortunate here, as complex numbers can now contain exact fractions, and
+   we'd want a nearest integer generic function for those)"
   [z]
-  (cond (complex? z)
-        (complex
-         (Math/round ^Float (real z))
-         (Math/round ^Float (imaginary z)))
+  (cond (complex? z) (complex (-> z real u/double Math/round u/int)
+                              (-> z imaginary u/double Math/round u/int))
         (v/native-integral? z) z
-        :else (Math/round (double z))))
-
-(defn gaussian-integer?
-  "Returns true if `z` is a [Gaussian
-  integer](https://en.wikipedia.org/wiki/Gaussian_integer), i.e., a complex entry
-  with integral real and imaginary components.
-
-  [[gaussian-integer?]] will return true if the real and imaginary components
-  are within `epsilon` of integral values. See [[value/almost-integral?]] for
-  details."
-  [z]
-  (if (complex? z)
-    (and (g/almost-integral? (real z))
-         (g/almost-integral? (imaginary z)))
-    (and (v/real? z)
-         (g/almost-integral? z))))
+        :else (Math/round (u/double z))))
 
 ;; ## Complex GCD
 
@@ -127,24 +112,15 @@
   (cond (g/zero? l) r
         (g/zero? r) l
         (v/= l r)   (abs-real l)
-        (not (or (gaussian-integer? l)
-                 (gaussian-integer? r)))
-        (u/illegal "gcd can only be computed for gaussian integers, but
-        both arguments were not.")
+        (not (and (g/exact? l)
+                  (g/exact? r))) (u/illegal "gcd can only be computed for gaussian integers.")
 
-        (not (gaussian-integer? l))
-        (u/illegal "gcd can only be computed for gaussian integers, but first
-        argument was not.")
-
-        (not (gaussian-integer? r))
-        (u/illegal "gcd can only be computed for gaussian integers, but second
-        argument was not.")
-
-        :else (let [[l r] (if (< (g/magnitude l)
+        :else (let [[l r] (if (> (g/magnitude l)
                                  (g/magnitude r))
                             [l r] [r l])]
-                (loop [a (round l)
-                       b (round r)]
+                (loop [a l
+                       b r]
+                  (println "gcd" a b (if (g/zero? b) "done" [(g/div a b) (round (g/div a b))]))
                   (if (g/zero? b)
                     (abs-real a)
                     (recur b (g/sub a (g/mul (round (g/div a b)) b))))))))
@@ -204,7 +180,7 @@
 
 (defmethod g/add [::complex ::complex] [a b] (c/add a b))
 ;; XXX consider making these methods in impl to avoid an allocation
-(defmethod g/add [::complex ::v/real] [a n] (c/add a (complex n)))x
+(defmethod g/add [::complex ::v/real] [a n] (c/add a (complex n)))
 (defmethod g/add [::v/real ::complex] [n a] (c/add (complex n) a))
 
 (defmethod g/sub [::complex ::complex] [a b] (c/sub a b))
--- a/src/emmy/complex.cljc
+++ b/src/emmy/complex.cljc
@@ -106,12 +106,12 @@
   or (* -1 z)), whichever number has a positive real component."
   [z]
   (cond (complex? z)
-        (if (neg? (real z))
+        (if (g/negative? (real z))
           (g/negate z)
           z)
 
         (v/real? z)
-        (Math/abs ^double (u/double z))
+        (g/abs z)
 
         :else (u/illegal "not supported!")))
 
@@ -154,9 +154,9 @@
 (defmethod g/one? [::complex] [c] (and (g/one? (real c))
                                        (g/zero? (imaginary c))))
 (defmethod g/identity? [::complex] [c] (g/one? c))
-(defmethod g/zero-like [::complex] [_] ZERO)
-(defmethod g/one-like [::complex] [_] ONE)
-(defmethod g/identity-like [::complex] [_] ONE)
+(defmethod g/zero-like [::complex] [c] (c/zero-like c))
+(defmethod g/one-like [::complex] [c] (c/one-like c))
+(defmethod g/identity-like [::complex] [c] (c/one-like c))
 (defmethod g/freeze [::complex] [c]
   (let [re (real c)
         im (imaginary c)]
@@ -195,16 +195,16 @@
 
 (defmethod v/= [::complex ::complex] [a b] (c/equal? a b))
 (defmethod v/= [::complex ::v/real] [a n]
-  (and (zero? (imaginary a))
+  (and (g/zero? (imaginary a))
        (v/= (real a) n)))
 
 (defmethod v/= [::v/real ::complex] [n a]
-  (and (zero? (imaginary a))
+  (and (g/zero? (imaginary a))
        (v/= n (real a))))
 
 (defmethod g/add [::complex ::complex] [a b] (c/add a b))
 ;; XXX consider making these methods in impl to avoid an allocation
-(defmethod g/add [::complex ::v/real] [a n] (c/add a (complex n)))
+(defmethod g/add [::complex ::v/real] [a n] (c/add a (complex n)))x
 (defmethod g/add [::v/real ::complex] [n a] (c/add (complex n) a))
 
 (defmethod g/sub [::complex ::complex] [a b] (c/sub a b))
--- a/src/emmy/complex.cljc
+++ b/src/emmy/complex.cljc
@@ -62,14 +62,6 @@
   [z]
   (c/imaginary z))
 
-;; ## Type Extensions
-
-#?(:clj
-   (defmethod print-method emmy.complex.impl.Complex [v ^java.io.Writer w]
-     (.write w (str "#emmy/complex "
-                    [(c/real v)
-                     (c/imaginary v)]))))
-
 ;; ## Gaussian Integers
 
 (defn round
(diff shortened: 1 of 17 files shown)


littleredcomputer:
I am beginning to wonder if the philosophy of `complex.js` is exactly what we want in Clojure. In that source library, explicit pains are taken to construct the real part and imaginary part of the results separately, (perhaps) so that allocation is avoided and (more likely) that as much significance as possible is preserved. But most of the complex operations can be built up out of a few primitives like arithmetic, exp and log. The existing implementations are a bit intricate here. More so than (say) the C++ implementation in <complex>. Consider `pow`: our friends in C++ are content with 

```C++
  template<typename _Tp>
    inline complex<_Tp>
    __complex_pow(const complex<_Tp>& __x, const complex<_Tp>& __y)
    { return __x == _Tp() ? _Tp() : std::exp(__y * std::log(__x)); }
```

littleredcomputer:
Another thought is that the complex vs. "impl" divide doesn't really get us much because impl needs generic and value, so we could fold this together (for example, by putting the implementations right into the defmethod statement)

test/emmy/generators.cljc
```diff
@@ -70,6 +71,15 @@
                               :min excluded-upper
                               :max max})])))
 
+(defn reasonable-complex
```
sritchie, line 74: nice

test/emmy/ratio_test.cljc
```diff
@@ -141,8 +141,8 @@
            (g/sqrt #emmy/ratio 9/4))
         "Ratios should stay exact if the numerator and denominator are exact.")
 
-    (is (= #emmy/complex "0+1.5i"
-           (g/sqrt #emmy/ratio -9/4))
+    (is (= #emmy/complex [0 #emmy/ratio 3/2]
```
sritchie, line 144: that's nice!

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,680 @@
...
+       (cons 0.)         ;; cons 0 and reverse facilitate evaluation with Horner's method
+       (reverse))
+  ;; => (4.779477332387385E-14 -1.147074559772972E-11 2.08767569878681E-9 -2.755731922398589E-7
+  ;;     2.48015873015873E-5 -0.001388888888888889 0.04166666666666667 -0.5 0.0)
+  ;; We copy this sequence, in order to avoid the dependency complex -> series,
+  ;; which creates a cycle that would be difficult to break, and it's clear in
+  ;; this case that complex is lower in the abstraction hierarchy than series.
+  )
+
+(def ^:private cos-1-square-terms
```
sritchie, line 101: you might make this `:no-doc` so you can use it from other test namespaces, and then put the above into the series tests, with a note here pointing there?

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,616 @@
...
+      :cljs [IEquiv
+             (-equiv [a b] (equal? a b))
+
+             IPrintWithWriter
+             (-pr-writer
+              [z writer _]
+              (write-all
+               writer
+               "#emmy/complex "
+               (str [(.-re z) (.-im z)])))]))
```
sritchie, line 43: Adding this should get you the same feature for the clj side if it's not somewhere else already:

```clojure
#?(:clj
   (defmethod print-method Complex
     [^Complex z ^java.io.Writer w]
     (.write w "#emmy/complex ")
     (.write w (pr-str [(.-re z) (.-im z)]))))
```

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,616 @@
...
+  [^Complex a b]
+  (cond (instance? Complex b)
+        (and (v/= (.-re a) (.-re b))
+             (v/= (.-im a) (.-im b)))
+
+        (v/real? b)
+        (and (g/zero? (.-im a))
+             (v/= (.-re a) b))
+
+        :else (v/= a b)))
```
sritchie, line 67: will this ever bottom out at `equals` and recurse back to `equal?`, causing a loop?

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,616 @@
...
+  (.-im z))
+
+(defn abs
+  "Calculate the magnitude of the complex number"
+  [^Complex z]
+  (let [x (.-re z)
+        y (.-im z)
+        a (g/abs x)
+        b (g/abs y)]
+    (cond (and (< a 3000) (< b 3000))
```
sritchie, line 114: can we document this? presumably this is to kill overflow but I'd love a note about the choice, if you understand why.

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,616 @@
...
+           :else (let [a (g// a 2)
+                       b (g// b 2)]
+                   (g/+ LN2 (g// (g/log (g/+ (g/* a a) (g/* b b))) 2)))))))
+
+(def ^:private
+  complex-re
+  "Regular expression used to parse complex numbers"
+  #"([+-]?\d+(\.\d*)?([Ee][+-]?\d+)?)(\s?([+-])?\s?([+-]?\d+(\.\d*)?([Ee][+-]?\d+)?)[Ii])?")
+
+(comment
```
sritchie, line 143: move to tests?

src/emmy/complex/impl.cljc
```diff
@@ -0,0 +1,616 @@
...
+  (re-matches complex-re "0+3i")
+
+  )
+
+(defn parse
+  "Parse a complex number. We expect one or two floating point numbers.
+  If two, they must be separated by a sign (perhaps surrounded by at most
+  one space, the second number followed by I or i. Example: 1.2-3.4i)"
+  [s]
+  (if-let [[_ re re-frac re-expt _ sign im im-frac im-expt] (re-matches complex-re s)]
```
sritchie, line 154: nice!

src/emmy/value.cljc
```diff
@@ -92,13 +90,13 @@
   [x]
   #?(:clj
      (or (instance? Number x)
-         (instance? Complex x))
+         (core/= (kind x) :emmy.complex/complex))
      :cljs (or (cljs.core/number? x)
                (core/= "bigint" (goog/typeOf x))
                (instance? Fraction x)
                (instance? goog.math.Integer x)
                (instance? goog.math.Long x)
```
sritchie, line 98: is it significantly cheaper to check `(instance? emmy.complex.Complex x)`? just curious if kind is expensive here.

src/emmy/complex.cljc
```diff
@@ -7,167 +7,82 @@
...
-     (-pr-writer [x writer _]
-       (write-all
-        writer
-        "#emmy/complex "
-        (str [(obj/get x "re")
-              (obj/get x "im")])))))
-
 #?(:clj
-   (defmethod print-method Complex [^Complex v ^java.io.Writer w]
+   (defmethod print-method emmy.complex.impl.Complex [v ^java.io.Writer w]
```
sritchie, line 68: ah, I see you have this here already. I might recommend moving it next to the type, so if someone only requires `emmy.complex.impl` they get this behavior too.

src/emmy/complex.cljc
```diff
@@ -249,128 +152,92 @@
...
+(defmethod g/acoth [::complex] [z] (c/acoth z))
+
+(defmethod g/floor [::complex] [z] (c/floor z))
+(defmethod g/ceiling [::complex] [z] (c/ceil z))
+
+(defmethod g/cosh [::complex] [z] (c/cosh z))
+(defmethod g/sinh [::complex] [z] (c/sinh z))
+(defmethod g/tanh [::complex] [z] (c/tanh z))
+(defmethod g/sech [::complex] [z] (c/sech z))
+(defmethod g/csch [::complex] [z] (c/csch z))
```
sritchie, line 239: so nice to see these simplified!

test/emmy/complex_test.cljc
```diff
@@ -377,67 +394,195 @@
...
+              (is (near z (g/acsch (g/csch z)))))
+    (checking "csch ∘ acsch = id" 100 [z (sg/reasonable-complex)]
+              (is (near z (g/csch (g/acsch z)))))
+    (checking "asech ∘ sech = id" 100 [z (sg/reasonable-complex)]
+              (is (near (right-half-plane z) (g/asech (g/sech z)))))
+    (checking "sech ∘ asech = id" 100 [z (sg/reasonable-complex)]
+              (is (near z (g/sech (g/asech z)))))
+    (checking "acoth ∘ coth = id" 100 [z (sg/reasonable-complex)]
+              (is (near z (g/acoth (g/coth z)))))
+    (checking "coth ∘ acoth = id" 100 [z (sg/reasonable-complex)]
```
sritchie, line 529: so good!!

sritchie (commented):
Looks great, just a few comments, and then let's do it.

src/emmy/value.cljc
littleredcomputer: I would have done that except you need to use the type in `impl` which felt like cheating. Let me see if the time difference is significant. If it is, then I agree that we should choose the quicker thing. Alternately, we could make the protocol INumerical handle this?

src/emmy/complex.cljc
littleredcomputer: I agree/fixing

src/emmy/complex/impl.cljc
littleredcomputer: deleted, ty

src/emmy/value.cljc
littleredcomputer:
```Clojure
(let [c (emmy.complex/complex 1 2)]
  (time
   (dotimes [_ 100000]
     (assert (= :emmy.complex/complex (emmy.value/kind c))))))
"Elapsed time: 8.2825 msecs"

(let [c (emmy.complex/complex 1 2)]
  (time
   (dotimes [_ 100000]
     (assert (instance? Complex c)))))
"Elapsed time: 1.419541 msecs"
```

littleredcomputer: ∴ fixing

src/emmy/complex/impl.cljc
littleredcomputer: Good question. I needed (I think) a way to allow the specializations for (e.g.) letting a complex number `=` a quaternion where the 1,i parts match and the j,k parts are zero. I'll look in to this

src/emmy/value.cljc
littleredcomputer: now we have the dependency problem again, which `kind` solved for us... maybe I'll leave it alone for now, and we could think of another way to do this

littleredcomputer: I spent a little more time on this. Yes, `kind` is about 3x more expensive than `instance?` for this test. But we can't import `Complex` into `value.cljc`. There's one way around this: a marker interface. We could use something like `(instance? emmy.value.Numerical v)` where instead of Numerical having a method we call, we use it purely as a marker interface, as something that should be thought of as part of the numeric tower without being known to `value`. We could look at what  `Numerical` is used for and try to make `numerical?` equivalent to either being part of the known numeric tower or possessing the marker interface.

littleredcomputer: OK: I added the marker interface INumericTower.

src/emmy/complex/impl.cljc
littleredcomputer: I have replaced the implementation with one cribbed from the GNU ISO C++ stdlib. That implementation pre-divides by the greater of the components and then corrects the result. It's much clearer what's going on.

src/emmy/complex/impl.cljc
littleredcomputer: I have imitated the implementation given for quaternions now, so there's no scary recursion possible.

littleredcomputer: This turned out to be a rabbit hole. A quaternion knows how to tell whether it is equal to a complex number, and it installs a method for v/= for (quaternion, complex) as well as (complex, quaternion). This is fine. 

There's a test in quaterion_test:
```clojure
(is (= i**i
               (g/expt q/I q/I)
               (g/expt q/J q/J)
               (g/expt q/K q/K))
            "i^i matches quaternion implementation for each of the components.")
```
wherein we call ordinary `=` on one complex and 3 quaternion values. How does it work in Clojure, since this is `clojure.core/=`? We see [here](https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/jvm/clojure/lang/Util.java#L31) that if either of the operands are `IPersistentCollection`s, we go [here](https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/jvm/clojure/lang/Util.java#L122), which solves the riddle: Quaternions are persistent collections, and while Complex numbers are not, the equiv will use the Quaternion implementation to finish the job. This doesn't work in ClojureScript, so we can solve this problem in one of a few ways: (1) Make the unit test use v/= (my preference), (2) Add the original fall-through logic you have just questioned above (the _status quo ante_), (3) equip Complex with enough behaviors to get this to work the way it does in Clojure (overkill IMO since the swapped comparison behavior in the persistent collection case I consider surprising). 

For the present, I'm going to make it so that the CLJS implementation does what it used to do, to keep the PR single-issue.

sritchie:
🎉

PR #154, opened 2023-11-21
Basic reverse-mode AD implementation [in progress]

sritchie:
Notes from my attempt at this a while back...

This is modeled after the approach described in the [dysvunctional language docs](https://github.com/axch/dysvunctional-language/blob/master/dvl/doc/ad.txt). This works now for `R^n => R` functions! It needs some work to handle nesting and all of that...

@littleredcomputer , I also took a try at an approach to share the derivatives between the `tape` and `differential` namespaces without actually making a `lift` implementation that lifts BOTH. The derivatives should certainly live somewhere, so we don't have to inline them.

Making progress!

## TODO

- [x] share derivative implementations (handled by #253)
- [ ] call the fn that calculates reverse mode a [`pullback`](https://www.juliadiff.org/ChainRulesCore.jl/stable/#The-propagators:-pushforward-and-pullback)?
- [ ] test that this can handle nesting. I'll almost certainly have to write some function to pull the tag of some differential up past a tape cell.... maybe?
- [ ] handle structural outputs too. Can I reuse my protocol definitions for `extract-tangent`, or do I have to redo everything?
- [ ] beef up the topological sort. here are some examples: [Kahn's](https://gist.github.com/alandipert/1263783) from @alandipert (described [on wiki](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm). Note that we almost certainly want to break ties lexicographically using the gensyms (OR the active tag stack!), look into this.
- [ ] note the idea from ChainRulesCore.jl about thunking. If you want to potentially not use ALL of the derivatives, then you might use a clojure `delay` to wrap some of the partials that take a bit of work. Look at the [julia docs](http://www.juliadiff.org/ChainRulesCore.jl/v0.6/api.html#ChainRulesCore.Thunk) to see how they build this in. Would we have to extend the generic arithmetic to delays?

## NOTES:
- Based on my investigations in #262 , we need to take some care there. We have some code to check if a `differential?` is incoming; that is obviously going to fail for reverse mode.
- the tie-breaking is only going to work if derivative itself uses `with-active-tag`, so let GJS know that we do that too.

### DIFF
--- a/src/emmy/tape.cljc
+++ b/src/emmy/tape.cljc
@@ -1,0 +1,352 @@
+#_"SPDX-License-Identifier: GPL-3.0";;
+
+(ns emmy.tape
+  "Basic machinery for reverse-mode AD! Let's see how far we get."
+  (:require [emmy.differential :as d]
+            [emmy.function :as f]
+            [emmy.generic :as g]
+            [emmy.structure :as s]
+            [emmy.util :as u]
+            [emmy.value :as v]))
+
+;; Start with a port of the DVL implementation.
+
+(derive ::tape ::f/cofunction)
+
+(deftype TapeCell [tag id primal in->partial]
+  v/Value
+  (zero? [_]
+    (v/zero? primal))
+
+  (one? [_]
+    (and (v/one? primal)
+         (empty? in->partial)))
+
+  (identity? [_]
+    (and (v/identity? primal)
+         (empty? in->partial)))
+
+  (zero-like [_] 0)
+  (one-like [_] 1)
+  (identity-like [_] 1)
+  (freeze [_]
+    `[~'TapeCell ~tag ~id ~(v/freeze primal)
+      ~(mapv (fn [[k v]] [(v/freeze k) (v/freeze v)])
+             in->partial)])
+  (exact? [_] false)
+
+  (kind [_] ::tape))
+
+;; These come from stdlib/ad-structures.dvl.
+
+(defn tape-tag [^TapeCell tape] (.-tag tape))
+(defn tape-id [^TapeCell tape] (.-id tape))
+(defn tape-primal [^TapeCell tape] (.-primal tape))
+(defn tape-partials [^TapeCell tape] (.-in->partial tape))
+
+;; TODO do we want the "epsilon" or "tag" to be a protocol? no way...
+
+(defn tape? [x]
+  (instance? TapeCell x))
+
+(defn make
+  "Make a `TapeCell` instance with a unique ID."
+  ([tag primal]
+   (make tag primal []))
+  ([tag primal partials]
+   (->TapeCell tag (gensym) primal partials)))
+
+(defn primal-part
+  "TODO THIS maybe could be a protocol if indeed the other implementations fall
+  down with it. This is a flag about how we're supposed to respond when FORWARD
+  mode tries to get our primal - do not give it up!!
+
+  NOTE from dvl... This expects that primal is always called with the outermost
+  available epsilon, and that bundles and tape cells are properly nested."
+  [x _tag]
+  x)
+
+(defn- reverse-primal
+  "This assumes that epsilon is greater than or equal to the perturbation of
+  thing, and is an epsilon associated with reverse mode."
+  [x tag]
+  (if (and (tape? x) (= tag (tape-tag x)))
+    (tape-primal x)
+    x))
+
+(defn primal*
+  "TODO this is also in ad-structures. This goes ahead and descends deep and
+  returns the ACTUAL primal part of either the tape or the differential... or ID
+  if it's neither of these container types."
+  [thing]
+  (if (tape? thing)
+    (primal* thing)
+    thing))
+
+;; TODO BOOM we actually... do we need to be more careful about extracting the
+;; appropriate primal part? I think the deal is that we get primals out of
+;; bundles, and just straight-up return them if they are tapes.
+
+(defn tangent-part
+  "TODO so weird, we ALSO only return tangent part here... oh. That is probably
+  because with proper nesting, if we're on the OUTSIDE then..."
+  [_x _tag] 0)
+
+(defn tag-of
+  "TODO redo this and make it so it doesn't HAVE to have a tag; but right now
+  we're returning 0 since it'll be less than all tags.
+
+  NOTE this is sort of like max-order-tag; but everything is not flattened out."
+  [x]
+  (if (tape? x)
+    (tape-tag x)
+    0))
+
+(comment
+  ;; ALSO in primal-part we have this tag hiding thing. So looks like that does
+  ;; not have to change.
+  (hide-gensym-in-procedure epsilon
+                            (lambda (x)
+                                    (primal epsilon (thing x)))))
+
+(defprotocol ITapify
+  (-tapify [this tag]))
+
+#_
+(comment
+  "TODO Do this later!"
+  (extend-protocol ITapify ...))
+
+(defn tapify
+  "Down to business! This should be a protocol method."
+  [x tag]
+  (cond (v/scalar? x)   (make tag x [])
+
+        (s/structure? x) (s/mapr #(tapify % tag) x)
+        (f/function? x)  (u/illegal "Can't do this yet.")
+
+        :else x))
+
+(defn reverse-phase
+  "sensitivities is a map... nodes is a sequence."
+  [nodes sensitivities]
+  (if (empty? nodes)
+    sensitivities
+    ;; Since you're going in topological sort order, when you reach
+    ;; a node you know you are done updating its sensitivity.
+    (let [sensitivity (get sensitivities (tape-id (first nodes)))]
+      (loop [sensitivities sensitivities
+             partials      (tape-partials (first nodes))]
+        (if (empty? partials)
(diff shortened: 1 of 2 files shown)


src/emmy/calculus/derivative.cljc
```diff
@@ -441,17 +442,17 @@
...
 
 (def D
-  "Derivative operator. Takes some function `f` and returns a function
-  whose value at some point can multiply an increment in the arguments, to
-  produce the best linear estimate of the increment in the function value.
+  "Derivative operator. Takes some function `f` and returns a function whose value
+  at some point can multiply an increment in the arguments to produce the best
+  linear estimate of the increment in the function value.
 
-  For univariate functions, [[D]] computes a derivative. For vector-valued
```
littleredcomputer, line 448: It's a pedantic point whether a row vector is a Jacobian or not, but I do think that univariate was correct here. A scalar valued function of 2 variables gives you a down tuple when differentiated, and a univariate vector valued function gives you an up tuple, which I think of as a derivative because you can write it as $f'$

src/emmy/calculus/derivative.cljc
```diff
@@ -441,17 +442,17 @@
...
+  For scalar-valued functions, [[D]] computes a derivative. For vector-valued
   functions, [[D]] computes
   the [Jacobian](https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant)
   of `f`.
 
-  The related [[Grad]] returns a function that produces a structure of the
-  opposite orientation as [[D]]. Both of these functions use forward-mode
+  The related [[emmy.env/Grad]] returns a function that produces a structure of
+  the opposite orientation as [[D]]. Both of these functions use forward-mode
```
littleredcomputer, line 455: I think of Grad and D as the same thing, except Grad doesn't reverse tuple orientation because nobody learns about that in their first class in calculus of several variables, but that's years of exposure to SICM. I think we should keep that behavior, though, since in the ML community people expect to dot the gradient with the increment to form the directional derivative; so I think both choices {grad vs. D}, {forward vs. backward} could be made independently, IIUC?

src/emmy/tape.cljc
```diff
@@ -0,0 +1,706 @@
...
+;;
+;; First, wrap all inputs to the function in an instance of a type called
+;; a [[TapeCell]]. This type holds
+;;
+;;   - a "tag", used as in [[emmy.differential]] to prevent confusion
+;;     between [[TapeCell]] instances created for different nested runs of
+;;     differentation
+;;   - a unique ID, used to de-duplicate work in the reverse pass
+;;   - the "primal value" wrapped by this [[TapeCell]] instance
+;;   - A map of the [[TapeCell]]'s inputs => the partial derivative of the
```
littleredcomputer, line 56: - this is not known initially, right? so that the process will assoc the partials as they become known?
- the tags are not unique among tape cells, but the unique IDs are?
- would `identical?` serve to dedup the work in the backward pass? (maybe I'll learn the answer to that as I read along, but since the field is populated by gensym, it would appear that the only property two gensyms have that is useful is not being identical, and no two TapeCells can be identical either)

src/emmy/tape.cljc
```diff
@@ -0,0 +1,706 @@
...
+;; out of primitives that are overloaded to deal with [[TapeCell]] instances.
+;;
+;; A primitive `f` of `n` arguments, given [[TapeCell]] inputs, should return a
+;; new [[TapeCell]] of the form:
+;;
+;; ```clojure
+;; {:tag <input tag>
+;;  :id <fresh ID>
+;;  :primal (f a b c ...)
+;;  :in->partials {<tape-a> <(((partial 0) f) a b c ...)>
```
littleredcomputer, line 69: `<tape-a>` etc. are references to other tape cells, pointing "in" to this one? Could they store the partials themselves I wonder

src/emmy/tape.cljc
```diff
@@ -0,0 +1,706 @@
...
+  "More permissive version of [[tape-tag]] that returns ##-Inf, the 'least
+  possible tag', when passed a non-[[TapeCell]] instance.
+
+  TODO this will need to be extended to
+  handle [[emmy.differential/Differential]] instances when these namespaces
+  merge."
+  [x]
+  (if (tape? x)
+    (tape-tag x)
+    0))
```
littleredcomputer, line 203: $0\ne -\infty$

littleredcomputer:
Having read this twice I think it's a beautifully compact implementation of the idea. I'm eager to try to understand how it works by printing out the tape structure before the reverse process.

src/emmy/tape.cljc
sritchie: - @littleredcomputer for inputs to the function, we start with an empty map; the other place that these `TapeCell`s are made is in `lift-1` and `lift-2`, and there the partials are known right at creation.

- the tags are there to distinguish between nested runs of differentiation, as in `(gradient (gradient f))`, and the IDs are there because a `TapeCell` might show up as an input to many other `TapeCell` instances, like with `z` in

```clojure
(fn [x y]
  (let [z (* x y)
        a (cos z)
        b (sin z)
        c (tan z)]
    (* a b c)))
```

- ANDDDDD I think you are right here that `identical?` would serve! I don't think we ever modify a `TapeCell` after it's created...

src/emmy/calculus/derivative.cljc
sritchie: Yeah, I agree, these should be independently tuneable

src/emmy/tape.cljc
```diff
@@ -0,0 +1,352 @@
...
+      [seen (cons node sorted)])))
+
+(defn compute-visiting-order* [nodes seen sorted]
+  (if (empty? nodes)
+    [seen sorted]
+    (let [[seen sorted] (compute-visiting-order
+                         (first nodes) seen sorted)]
+      (compute-visiting-order* (rest nodes) seen sorted))))
+
+(defn interpret [thing tag sensitivities]
```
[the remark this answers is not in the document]
sritchie: yeah, that's great

src/emmy/calculus/derivative.cljc
```diff
@@ -396,7 +397,7 @@
          df (d f)
          df* (d (fn [args] (apply f args)))]
      (-> (fn
-           ([] (constantly 0))
```
sritchie, line 399: @littleredcomputer this was a bug! for a no-arg `f`, `((D f))` should return 0, not another function.

src/emmy/calculus/derivative.cljc
```diff
@@ -292,7 +293,7 @@
    (let [entry (get-in structure path)]
      (deep-partial f structure path entry)))
   ([f structure path entry]
-   (if (v/numerical? entry)
+   (if (v/scalar? entry)
```
sritchie, line 296: @littleredcomputer this change leaves us free to make `numerical?` return `false` for our differential AND tape instances.

test/emmy/tape_test.cljc
```diff
@@ -0,0 +1,495 @@
...
+
+  (testing "exercise some of the lifted fns by comparing them to numeric
+  derivatives."
+    (let [f (fn [x]
+              (g/+ (g/* (g/cos x) (g/sin x))
+                   (g/+ (g/sin x) (g/expt x 2))
+                   (g/+ (g/sin x) x)
+                   (g/log (g/abs x))))
+          Df         (t/gradient f)
+          Df-numeric (D-numeric f)]
```
littleredcomputer, line 281: is the idea here that you don't want to pass when forward and backward are both wrong?

test/emmy/tape_test.cljc
```diff
@@ -0,0 +1,495 @@
...
+                      (g (g/+ a offset)))))
+          f-hat ((t/gradient shift) 3)]
+      (is (= (g/exp 8) ((f-hat g/exp) 5))
+          "Nothing tough expected in this case.")
+
+      ;; We'll update this once we fix the amazing bug for reverse-mode.
+      (is (= 0 ((f-hat (f-hat g/exp)) 5))
+          "This is the amazing bug, and SHOULD actually equal (g/exp 11)."))))
+
+(deftest sinc-etc-tests
```
littleredcomputer, line 307: An idea (for the future!) might be to probabilistically generate smallish expression trees out of safely-differentiable components (+, -, *, exp) and compare forward, backward, numerical results

test/emmy/generators.cljc
```diff
@@ -312,6 +314,36 @@
                :else (d/from-terms [(#'d/make-term [] primal)
                                     (#'d/make-term [0] 1)])))))))
 
+;; ## TapeCell
```
littleredcomputer, line 317: oh wow, you did it! how did I miss that. Great minds think alike

src/emmy/tape.cljc
```diff
@@ -246,24 +305,104 @@
...
+     false)))
+
+(defn compare
+  "Comparator that compares [[Differential]] instances with each other or
+  non-differentials using only the [[finite-part]] of each instance. Matches the
+  response of [[equiv]].
+
+  Acts as [[emmy.value/compare]] for non-[[TapeCell]]s."
+  [a b]
+  (v/compare
```
littleredcomputer, line 384: Haskell calls this pattern `on v/compare tape-primal`, usually written infix. I don't know how RH could resist that one after `just`

PR #155, opened 2023-11-30
Replace bigfraction.js with CLJS code

littleredcomputer:
A minimal set of arbitrary precision rationals is built on a pair of native JS BigInts. Rather than manage a general fraction field over some Euclidean domain, it was decided to produce something very similar to Clojure's native Ratio type.

Unusually for an Emmy file, the implementation is ClojureScript only; the main technicality in this library is preserving the BigInt-compatibility of the arithmetic operations done.

### DIFF
--- a/src/emmy/bigfraction.cljs
+++ b/src/emmy/bigfraction.cljs
@@ -115,10 +115,13 @@
   (if (< a 0) (* -ONE a) a))
 
 (defn integer->
+  "Create a fraction with unit denominator."
   [n]
   (->Fraction (js/BigInt n) ONE))
 
 (defn ->real
+  "Coerce a fraction to real by performing the division
+   in the floating point domain"
   [^Fraction q]
   (/ (js/Number (.-n q)) (js/Number (.-d q))))
 
@@ -129,9 +132,9 @@
   (when (== ZERO b)
     (division-by-zero))
   (let [an (< a 0)
-        a (if an (* -ONE a) a)
+        a (bigint-abs a)
         bn (< b 0)
-        b (if bn (* -ONE b) b)
+        b (bigint-abs b)
         g (bigint-gcd a b)
         neg (not= an bn)
         abs_c (/ a g)
--- a/src/emmy/bigfraction.cljs
+++ b/src/emmy/bigfraction.cljs
@@ -1,751 +1,316 @@
-(ns bigfraction)
-
-(def C-ONE (js/BigInt 1))
-(def C-ZERO (js/BigInt 0))
-(def C-TEN (js/BigInt 10))
-(def C-TWO (js/BigInt 2))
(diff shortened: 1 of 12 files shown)


src/emmy/bigfraction.cljs
```diff
@@ -0,0 +1,319 @@
...
+(declare eq cmp ->real)
+
+(deftype Fraction [^js/BigInt n ^js/BigInt d]
+
+  Object
+  (valueOf [this] (->real this))
+  (toString [_] (str n "/" d))
+
+  ;; these are the remaining dependencies on value.
+  ;; Maybe they should be part of extend-type in the ratio library.
```
sritchie, line 32: I'd be okay with that! Have `ratio` require this and extend-type.

src/emmy/bigfraction.cljs
```diff
@@ -0,0 +1,319 @@
...
+        d (.-d y)]
+    (and (== a c) (== b d))))
+
+(defn bigint-gcd
+  "GCD assuming a and b are BigInts > 0"
+  [^js/BigInt a ^js/BigInt b]
+  (loop [a a
+         b b]
+    (if (== b ZERO) a
+        (recur b (core/mod a b)))))
```
sritchie, line 112: nice and core/mod works for bigint?

I think `js-mod` would be cheaper here and identical in behavior for positive numbers...

src/emmy/bigfraction.cljs
```diff
@@ -0,0 +1,319 @@
...
+
+(defn- bigint-abs [a]
```
sritchie, line 114: what do you think the original implementation gained by having the sign bit?

src/emmy/ratio.cljc
```diff
@@ -12,23 +12,20 @@
...
-            #?(:cljs [goog.object :as obj])
+            #?(:cljs [emmy.bigfraction :as bf])
             [emmy.generic :as g]
             [emmy.util :as u]
             [emmy.value :as v])
   #?(:clj (:import (clojure.lang Ratio))))
 
 (def ^:no-doc ratiotype
-  #?(:clj Ratio :cljs Fraction))
+  #?(:clj Ratio :cljs ::ratio))
```
sritchie, line 22: curious, why this vs `bf/Fraction`?

sritchie (approved):
LGTM, this is dense stuff! I'm glad we have so many tests...

sritchie:
Good to go except for maybe the `js-mod` change if that is faster

src/emmy/bigfraction.cljs
littleredcomputer: agree, this is an oversight/will fix. Initially I didn't do the work as cljs-only, but that turned out to be a liberating way to approach it instead of hoping that core/ did what I wanted in these cases

PR #156, opened 2023-12-08
Support nested forward/reverse mode

sritchie:
This PR:

  - Makes forward- and reverse-mode automatic differentiation compatible with
    each other, allowing for proper mixed-mode AD

  - Adds support for derivatives of literal functions in reverse-mode
  
  - Adds support for `emmy.operator/Operator`-shaped outputs in reverse-mode

In a future PR I want to add to / modify the `IPerturbed` protocol so that it's as supportive of reverse-mode as forward-mode. I'm leaving that to the future for now.


littleredcomputer (approved):

src/emmy/tape.cljc
```diff
@@ -243,15 +243,11 @@
 
 (defn tag-of
   "More permissive version of [[tape-tag]] that returns ##-Inf, the 'least
-  possible tag', when passed a non-[[TapeCell]] instance.
-
-  TODO this will need to be extended to
-  handle [[emmy.differential/Differential]] instances when these namespaces
-  merge."
+  possible tag', when passed a non-[[TapeCell]] instance."
```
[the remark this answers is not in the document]
sritchie: I chose "perturbation", what do you think?

src/emmy/calculus/derivative.cljc
```diff
@@ -531,12 +532,23 @@
...
+                                 (rec (d/primal x))
+                                 (rec (d/tangent x))
+                                 (d/tag x))
+
+                                (tape/tape? x)
+                                (tape/->TapeCell
+                                 (tape/tape-tag x)
+                                 (tape/tape-id x)
+                                 (rec (tape/tape-primal x))
+                                 (mapv (fn [[node partial]]
```
littleredcomputer, line 546: looks like the inner thing is `mapv` too. Does writing it that way feel less janky?  `[node partial]` is almost an instance of Dual, right, except that the tag belongs to the containing tape. Armed with that, you could unify this part with the part above, but (I say sincerely) it would be too clever. Therefore I don't think this is janky: it feels like about what we want to see here

src/emmy/tape.cljc
```diff
@@ -317,15 +307,65 @@
...
+  containing the tag and instance of [[emmy.differential/Dual]] or [[TapeCell]]
+  associated with the inner-most call to [[with-active-tag]] in the current call
+  stack.
+
+  If none of `dxs` has an active tag, returns `nil`."
+  ([& dxs]
+   (let [m (into {} (mapcat
+                     (fn [dx]
+                       (when-let [t (tag-of dx)]
+                         {t dx})))
```
littleredcomputer, line 343: I wonder if it is cheaper to write `[t dx]`. However you choose to write it, it's cool (and I do think that `{t dx}` is a reasonable way to write the lone binding

src/emmy/tape.cljc
```diff
@@ -699,43 +763,37 @@
...
+  `f`, [[TapeCell]] and [[emmy.differential/Dual]] instances.
 
   If called without `df:dx`, `df:dx` defaults to `(f :dfdx)`; this will return
   the derivative registered to a generic function defined
   with [[emmy.util.def/defgeneric]].
 
-  NOTE: `df:dx` has to ALREADY be able to handle [[TapeCell]] instances. The
-  best way to accomplish this is by building `df:dx` out of already-lifted
-  functions, and declaring them by forward reference if you need to."
+  NOTE: `df:dx` has to ALREADY be able to handle [[TapeCell]]
```
littleredcomputer, line 773: this makes sense; it's so sweet. keep the magic at the most primitive layer

src/emmy/tape.cljc
```diff
@@ -944,9 +1019,29 @@
 ;; Non-differentiable generic operations
 
 (defbinary v/= (by-primal v/=))
-(defunary g/zero? (by-primal g/zero?))
-(defunary g/one? (by-primal g/one?))
-(defunary g/identity? (by-primal g/identity?))
+(defunary g/zero?
```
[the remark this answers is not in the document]
littleredcomputer: Do we need `g/zero?` for tapes? If not, I'd leave it completely undefined: those things are kind of dangerous as you have discovered.

test/emmy/tape_test.cljc
```diff
@@ -11,6 +11,7 @@
             [emmy.generators :as sg]
             [emmy.generic :as g]
             [emmy.numerical.derivative :refer [D-numeric]]
+            [emmy.operator :as o]
```
littleredcomputer, line 14: I just had this random test idea, which I don't think is particularly valuable or anything, but thought it was cool: "the Jacobian of a permutation is the permutation matrix of that permutation", as in :

```
(defn f [a b c d e] (up d e c b a))
(def M ((D f) 'a 'b 'c 'd 'e))
;; M = (down (up 0 0 0 0 1) (up 0 0 0 1 0) (up 0 0 1 0 0) (up 1 0 0 0 0) (up 0 1 0 0 0))
;; and so:
 (* M [1 2 3 4 5])
;; => (up 4 5 3 2 1)

**Review by littleredcomputer — APPROVED** [review]

File: src/emmy/abstract/function.cljc [hunk 7]
```diff
@@ -262,15 +277,50 @@
...
+  [f primal-s tag]
+  (fn
+    ([] [])
+    ([partials]
+     (tape/make tag (apply f primal-s) partials))
+    ([partials [entry path _]]
+     (if (and (tape/tape? entry) (= tag (tape/tape-tag entry)))
+       (let [partial (literal-partial f path)]
+         (conj partials [entry (literal-apply partial primal-s)]))
+       partials))))
```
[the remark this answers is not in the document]
sritchie: this would happen in the case of nested `gradient` calls to a literal function, I just don't have those in the tests yet. In that case, the innermost tag wins and any tag with a different tape is treated as a scalar (i.e. partial derivative == 0, so we never add it to the map)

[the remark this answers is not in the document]
sritchie: I want to share tests between forward mode and reverse mode... when I do that this should get hit.

src/emmy/calculus/derivative.cljc
sritchie: this whole function is too clever, and my cleverness bit me here: https://github.com/mentat-collective/emmy/issues/168

sritchie: cleverness in trying to make this compatible with `D`, that is...

test/emmy/tape_test.cljc
sritchie: Added as a test!

src/emmy/tape.cljc
sritchie: your way is faster!!

PR #157, opened 2024-02-09
fix(extend-type): use bigint instead of BigInt

mhuebert:
I was seeing this warning on recent cljs:

```
WARNING: Extending an existing JavaScript type - use a different symbol name instead of js/BigInt
```

Using `bigint` instead of `js/BigInt` when extending types/protocols eliminates this warning. via @shaunlebron:

> The "function", "object", "array" symbols are just symbols that have meaning in the context of the `extend-type` macro, so nothing special syntactically. They just map internally to the type strings produced by `goog/typeOf`, used for protocol dispatch.

I verified that `(goog/typeOf (js/BigInt 1))` => "bigint"

Discussion: https://groups.google.com/g/clojurescript/c/MKEZ9CBU77o?pli=1

### DIFF
--- a/src/emmy/value.cljc
+++ b/src/emmy/value.cljc
@@ -242,7 +242,7 @@
            (= this (.valueOf other)))))))
 
 #?(:cljs
-   (extend-type js/BigInt
+   (extend-type bigint
      IHash
      (-hash [this] (hash (.toString this 16)))
 
@@ -288,7 +288,7 @@
            (garray/defaultCompare this other)
            (throw (js/Error. (str "Cannot compare " this " to " o))))))
 
-     js/BigInt
+     bigint
      (-compare [this o]
        (let [other (.valueOf o)]
          (if (real? other)
@@ -313,7 +313,7 @@
    ;; ClojureScript-specific implementations of Value.
    (do
      (extend-protocol Numerical
-       js/BigInt
+       bigint
        (numerical? [_] true)
 
        goog.math.Integer
@@ -323,7 +323,7 @@
        (numerical? [_] true))
 
      (extend-protocol IKind
-       js/BigInt
+       bigint
        (kind [_] js/BigInt)
 
        goog.math.Integer


borkdude:
This is new CLJS `1.11.132` behavior right?
For now configure clj-kondo with `{:linters {:unresolved-symbol [bigint]}}`
I'll fix it in the next release of clj-kondo (unfortunately just had one today already)

sritchie:
Thanks @borkdude !

sritchie:
Interesting, @mhuebert this seems to fail with the version of cljs used in the tests;

```
TypeError: Cannot read properties of undefined (reading 'prototype')
    at /home/runner/work/emmy/emmy/.shadow-cljs/builds/test/dev/out/cljs-runtime/emmy/value.cljc:245:4
    at global.SHADOW_IMPORT (/home/runner/work/emmy/emmy/target/main/node-tests.js:64:44)
    at /home/runner/work/emmy/emmy/target/main/node-tests.js:1582:1
    at Object.<anonymous> (/home/runner/work/emmy/emmy/target/main/node-tests.js:1933:3)
    at Module._compile (node:internal/modules/cjs/loader:1198:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
    at Module.load (node:internal/modules/cjs/loader:1076:32)
    at Function.Module._load (node:internal/modules/cjs/loader:911:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:22:47
Error: Process completed with exit code 1.
```

I wonder if this is not backwards compatible. I'm confused since based on that group post it should be...

borkdude:
I don't think it's backwards compatible, since the support for this was explicitly added in a new version of CLJS. There is no way to make prior versions of CLJS behave differently?

sritchie:
For sure, I just didn't know if the warning was new or the functionality for bigint.

sritchie:
@mhuebert , I'm going to close this so that we keep backwards compatibility - lmk and we can re-open if that's a problem. Thank you for finding this, let's do it when appropriate!

PR #159, opened 2024-04-08
Add fold-chain, convert numerical? calls to scalar? (1/N)

sritchie:
This PR:

  - Fixes `Differential`'s implementation of `emmy.value/numerical?` to always
    return `false`. The reason is that `numerical?` is used by `g/*` and friends
    to decide on simplifications like `(* <dx-with-1> x) => x`, which would lose
    the structure of `dx-with-1`. By returning false we avoid these
    simplifications.

  - Converts a number of `emmy.value/numerical?` calls to `emmy.value/scalar?`.
    The `numerical?` protocol method is used only in generic functions like
    `g/*` for deciding whether or not to apply numerical simplifications, like
    `(* x 1) => x`.

    Guarding on `v/scalar?` instead allows us to let in numbers, tapes and
    differentials, since these latter two are meant to WRAP numbers, but should
    not be subject to numerical simplifications.

  - Adds `emmy.structure/fold-chain` for performing a tree-like fold on
    structures, saving us work over the alternate pattern of `s/mapr`, `flatten`
    and `reduce`.


sritchie:
@littleredcomputer I feel like I finally understand `numerical?`... and I also feel like we probably should make differentials return `false`.

sritchie:
Equals no longer working in cljs with this change is surprising. I'm glad the tests catch this, but what is going on... `valueOf` should be all that matters and that did not change.

sritchie:
Honestly I think we should just delete `numerical?` and use `number?` instead.

littleredcomputer:
> @littleredcomputer I feel like I finally understand `numerical?`... and I also feel like we probably should make differentials return `false`.

We now realize I think that the name communicates almost nothing about the intent...I don't propose it for this PR, but maybe there's a better name: can we capture what is intended? It seems like it contains the notion of "being in the center" of the semigroup induced by `*`. 

Could it be that it's not `numerical?` that promises that `(= x (* dx1 x))`, but `one?`

> EDIT yes I think we should definitely make `false` the return value. Maybe there is a more efficient way than a protocol to implement `numerical?`...

src/emmy/abstract/function.cljc
```diff
@@ -139,6 +140,13 @@
    (defmethod print-method Function [^Function f ^java.io.Writer w]
      (.write w (.toString f))))
 
+#?(:clj
```
littleredcomputer, line 143: where is this used? Did we have output were functions were ugly in the ordinary case? (curiosity only: it can be hard in Clojure-land to find out who has intervened in a situation like this)

src/emmy/structure.cljc
```diff
@@ -694,6 +694,63 @@
...
+   {:s 3, :chain [1 0], :orientations [::s/down ::s/up]}
+   {:s 4, :chain [1 1], :orientations [::s/down ::s/up]}]
+  ```"
+  ([f s] (fold-chain f (f) f s))
+  ([f init s] (fold-chain f init f s))
+  ([f init present s]
+   (letfn [(walk [acc s chain orientations]
+             (if (structure? s)
+               (let [o (orientation s)]
+                 (reduce
```
littleredcomputer, line 742: Does this mean we have one pending `reduce` per layer of "rank" in the structure? Would this be more "transducey" if we did _not_ call reduce, but instead, _generated_ the lazy sequence of (e, chain, orientations) and let that feed into a (trans|re)duction supplied elsewhere?

src/emmy/abstract/function.cljc
littleredcomputer: nvm I see in the test below. Is it fair to say that this is a workaround for a bug in `pprint`? I would think that once we had implemented `toString` it would be honored equally by [p]print

test/emmy/structure_test.cljc
```diff
@@ -542,6 +557,20 @@
                           (s/down 'g 'h)))))
         "sumr uses g/+, so symbols etc can be added too."))
 
+  (testing "fold-chain"
```
littleredcomputer, line 560: I like it. I can easily imagine wanting a level-cutoff feature, but that's only because I've been in numpy-land recently and that way of thinking is fresh on my mind, so please don't do it :)

littleredcomputer (commented):
LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,28 @@
...
+  - Fixes `Differential`'s implementation of `emmy.value/numerical?` to always
+    return `false`. The reason is that `numerical?` is used by `g/*` and friends
+    to decide on simplifications like `(* <dx-with-1> x) => x`, which would lose
+    the structure of `dx-with-1`. By returning false we avoid these
+    simplifications.
+
+  - Converts a number of `emmy.value/numerical?` calls to `emmy.value/scalar?`.
+    The `numerical?` protocol method is used only in generic functions like
+    `g/*` for deciding whether or not to apply numerical simplifications, like
+    `(* x 1) => x`.
```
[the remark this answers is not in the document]
sritchie: Here's the implementation with `numerical?` in it:

```clj
(defn * [x y]
  (let [numx? (v/numerical? x)
        numy? (v/numerical? y)]
    (cond (and numx? (zero? x)) (zero-like y)
          (and numy? (zero? y)) (zero-like x)
          (and numx? (one? x)) y
          (and numy? (one? y)) x
          :else (mul x y))))
```

you're right, it's something for which `one?` is true. This has bitten us in the past, where we had `(g/* x 0)` with a vector-shaped `x`... `g/*` returned 0, throwing away the structure when it should have returned `x`. That's what prompted the (zero-like ...), I believe.

I am nervous that this in-line simplification is just not a good idea unless we KNOW we're dealing with a number...

src/emmy/differential.cljc
```diff
@@ -565,15 +565,12 @@
...
+  ;; A [[Differential]] has to respond `false` to all [[emmy.value/numerical?]]
+  ;; inquiries; if we didn't do this, then [[emmy.generic/*]] and friends would
+  ;; attempt to apply shortcuts like `(* x <dx-with-1>) => x`, stripping off
+  ;; the [[Differential]] identity of the result and ruining the derivative.
   v/Numerical
-  (numerical? [_]
-    (or (empty? terms)
-        (v/numerical?
-         (coefficient (nth terms 0)))))
+  (numerical? [_] false)
```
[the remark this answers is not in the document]
sritchie: It does... but it ALSO frustrates the simplification of `(* <1, dx> y) => y`, where `<1, dx>` is a dual with `1` in its primal slot.

I had fixed this before by making `(one? <1, dx>)` return false, but `(= 1 <1, dx>)` return true. That feels so fiddly and easy to screw up. 

I'd rather have the multiplication proceed, even when we could have saved a touch of work. wdyt?

src/emmy/mechanics/hamilton.cljc
```diff
@@ -53,9 +53,9 @@
   (and (s/up? s)
        (= (count s) 3)
        (let [[t q v] s]
-         (and (v/numerical? t)
-              (or (and (v/numerical? q)
-                       (v/numerical? v))
+         (and (v/scalar? t)
```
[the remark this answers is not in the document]
sritchie: Yeah, this came up during some test with a symbolic derivative, I believe. I can turn it off and run the tests to check if you like

src/emmy/structure.cljc
```diff
@@ -694,6 +694,63 @@
   [f & structures]
   (sum:r:l f structures))
 
+(defn fold-chain
```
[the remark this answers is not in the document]
sritchie: The next PR uses it to sum up all of the partial derivative contributions from a structure passed in to `literal-derivative`... right now we use `flatten` which I don't really like. Walking with `fold-chain` lets us avoid building another structure with `mapr` and then immediately flattening it (building yet another sequence) and reducing that sequence to get the sum.

I'll use a different fold function to collect all the pairs of `[input partial-input]` for the reverse mode forward pass in `literal-derivative`... I'll put that together after we get these in!

src/emmy/abstract/function.cljc
sritchie: yes! ugh, I should have written this down... i was getting a `LazySeq` output... but `pprint` was deferring to some other method than our `toStr` here. I should justify this...

src/emmy/structure.cljc
sritchie: yes I bet it would! the `fold` name works here because the accumulator is passed up through the levels, but your idea would let us use normal `transduce`, which would be great.

PR #163, opened 2024-04-11
Replace Differential implementation with simplified Dual (2/N)

sritchie:
This PR 

- replaces The `Differential` generalized dual and its term list algebra with a simplified `Dual` number type
- deletes `emmy.util.vector-set` and tests, as these are no longer used
- adds a `nil` implementation for `extract-tangent`, meaning that `nil`-valued functions now work with `D`

This new approach works because the `emmy.differential/*active-tags*` stack allows us to make sure that lifted binary operations always wrap their output in a new `Dual` with the tag assigned by the inner-most derivative call.


src/emmy/differential.cljc
```diff
@@ -746,284 +539,117 @@
...
+  via [[with-active-tag]] (i.e., the tag connected with the _innermost_ call
+  to [[with-active-tag]]).
+
+  If none of the tags are bound, returns `(apply max tags)`."
+  [& tags]
+  (or (some (apply hash-set tags)
+            *active-tags*)
+      (apply max tags)))
+
+(defn tag+perturbation
```
littleredcomputer, line 571: could you just return the dual number? it has the tag inside it already?

src/emmy/differential.cljc
```diff
@@ -1331,20 +960,30 @@
...
+           (g/zero? t)))))
+
+(defunary g/one? one?)
+(defunary g/identity? identity?)
+
+(defmethod g/zero-like [::dual] [_] 0)
+(defmethod g/one-like [::dual] [_] 1)
+(defmethod g/identity-like [::dual] [_] 1)
+(defmethod g/freeze [::dual] [d]
+  `[~'Dual
```
littleredcomputer, line 981: I swore a saw a dict-like implementation of this somewhere

littleredcomputer (approved):
Wow. This is absolutely phenomenal.

src/emmy/calculus/derivative.cljc
```diff
@@ -530,8 +531,11 @@
      (letfn [(process-term [term]
                (g/simplify
                 (s/mapr (fn rec [x]
-                          (if (d/differential? x)
-                            (d/map-coefficients rec x)
+                          (if (d/dual? x)
```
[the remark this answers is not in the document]
sritchie: This has always seemed like a hack to me... but maybe not. yes, it's like an `fmap` for sure, and it's what we do for `simplify`. We'll have a similar line to add for tapes. What do you think? Should we add something more general?

src/emmy/differential.cljc
sritchie: I had a `pprint` implementation for tapes that did that:

https://github.com/mentat-collective/emmy/blob/main/src/emmy/tape.cljc#L285-L296

but I didn't use it for `freeze`. Honestly I don't think these should ever make it out into a situation where we'd inspect the results of a freeze...

src/emmy/differential.cljc
sritchie: YES that is true... I added this on my scratch branch when I still had `Differential` and it was a touch more expensive to get the tag out, so I wanted to save the effort.

Can we look at this together when I massage this to be compatible with tapes too?

PR #165, opened 2024-04-15
Add ID back to tape, fix Alexey's Amazing Bug in reverse-mode

sritchie:
This PR:

- Fixes Alexey's Amazing Bug for our tape implementation
- Adds the `id` field back into `TapeCell`, required for the tag replacement machinery for fixing Alexey's Bug
- Adds an `emmy.differential/IPerturbed` implementation to TapeCell
- Fixes some old documentation referencing `:in->partials`, plural
- Updates `emmy.tape/tag-of` to return nil in case of a non-tape argument vs `##-Inf`, again preparation for forward/reverse interactions
- Adds support for partial derivatives via a `selectors` argument to `gradient`
- Fixes a bug in the `g/abs` implementation for `TapeCell`, where `(g/abs <tape>)` would return a positive primal
- Adds a `simplify` implementation for `TapeCell`

### DIFF
--- a/src/emmy/tape.cljc
+++ b/src/emmy/tape.cljc
@@ -60,6 +60,7 @@
 ;;   - a "tag", used as in [[emmy.differential]] to prevent confusion
 ;;     between [[TapeCell]] instances created for different nested runs of
 ;;     differentation
+;;   - a unique ID, used to de-duplicate work in the reverse pass
 ;;   - the "primal value" wrapped by this [[TapeCell]] instance
 ;;   - A vector of pairs of each [[TapeCell]] input paired with the partial
 ;;     derivative of the primal value with respect to that input
@@ -74,10 +75,11 @@
 ;; ```clojure
 ;; {:tag <input tag>
 ;;  :primal (f a b c ...)
-;;  :in->partials [[<tape-a> <(((partial 0) f) a b c ...)>]
-;;                 [<tape-b> <(((partial 1) f) a b c ...)>]
-;;                 [<tape-c> <(((partial 2) f) a b c ...)]>
-;;                 ...]
+;;  :id <fresh ID>
+;;  :in->partial [[<tape-a> <(((partial 0) f) a b c ...)>]
+;;                [<tape-b> <(((partial 1) f) a b c ...)>]
+;;                [<tape-c> <(((partial 2) f) a b c ...)]>
+;;                ...]
 ;; }
 ;; ```
 ;;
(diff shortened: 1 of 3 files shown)


src/emmy/tape.cljc
```diff
@@ -455,6 +495,32 @@
 ;; values like maps or quaternions. [[emmy.differential/extract-tangent]] does
 ;; this for forward-mode, I believe.
 
+(declare ->partials)
+
+(defn- ->partials-fn
```
sritchie, line 500: This is the solution to Alexey's Amazing Bug, along with the use of `tag+perturbation` in `lift-2` below. Notice how similar it is to `extract-tangent-fn`... these ideas should be merged, and IDEALLY the tapes could use `extract-tangent` too. Still thinking about this.

src/emmy/tape.cljc
```diff
@@ -477,13 +543,12 @@
...
 
-        ;; TODO this needs to handle perturbation confusion with tag
-        ;; replacement. Make something similar to extract-tangent-fn and test
-        ;; this so we don't allow an amazing bug.
-        (comp #(->partials % tag) output)
+        (v/scalar? output)
+        (->Completed {})
 
-        :else (->Completed {})))
+        :else (u/unsupported "Output not handled yet!")))
```
sritchie, line 551: I want to replace this `cond` with something better, but I'm still thinking about it...

src/emmy/tape.cljc
```diff
@@ -538,10 +603,9 @@
   computation at that leaf."
   [input output tag]
   (cond (and (tape? input) (= tag (tape-tag input)))
-        (extract output input)
+        (extract output (tape-id input))
 
         (s/structure? input)
-        ;; TODO if we don't care about flipping for gradient, then use mapr.
```
sritchie, line 544: let's hold the line and always return the opposite variance :)

src/emmy/tape.cljc
```diff
@@ -571,19 +635,30 @@
...
-     (let [tag       (d/fresh-tag)
-           inputs    (tapify x tag)
-           output    (f inputs)
-           completed (->partials output tag)]
-       (interpret inputs completed tag)))
-    ([x & more]
-     ((gradient (fn [args]
-                  (apply f args)))
-      (matrix/seq-> (cons x more))))))
+  ([f] (gradient f []))
```
sritchie, line 638: THis needs docs, and a test case for the partial derivative.

littleredcomputer (commented):
LGTM!

PR #169, opened 2024-04-23
Support all output types in reverse-mode

sritchie:
This PR:

  - Adds a new `mode` argument to `emmy.differential/extract-tangent`, allowing us to use this for reverse mode instead of `emmy.tape/->partials` (which is now gone)
  - replaces `emmy.tape/extract` with `emmy.differential/extract-id`, now part of the `IPerturbed` protocol.

We need these two passes because:

1. for structural outputs, we need to do a reverse-mode pass for each entry in the output. This creates a structure of `Completed` maps of `node => sensitivity`.
2. then, for each entry in a structural INPUT, we need to replace that entry with a copy of this output with that input's ID selected out


src/emmy/calculus/derivative.cljc
```diff
@@ -202,18 +202,19 @@
 ;; preserved through tag replacement and extraction.
 
 (extend-protocol d/IPerturbed
-
   MultiFn
   (perturbed? [_] false)
   (replace-tag [f old new] (replace-tag-fn f old new))
-  (extract-tangent [f tag] (extract-tangent-fn f tag))
+  (extract-tangent [f tag mode] (extract-tangent-fn f tag mode))
+  (extract-id [f id] (comp #(d/extract-id % id) f))
```
sritchie, line 209: it's kind of an `fmap` for most types... after your multimethod conversion, @littleredcomputer , do you feel like these should be multimethods too? Then we could have a `::functor` type that supports fmap?

src/emmy/calculus/derivative.cljc
```diff
@@ -258,7 +262,7 @@
     (let [tag    (d/fresh-tag)
           lifted (d/bundle-element x 1 tag)]
       (-> (d/with-active-tag tag f [lifted])
-          (d/extract-tangent tag)))))
+          (d/extract-tangent tag ::d/dual)))))
```
sritchie, line 265: I should probably bind this and `::tape/tape` to vars so there's no chance to mis-type these.

src/emmy/differential.cljc
```diff
@@ -337,9 +338,35 @@
...
 
-  (extract-tangent [this tag]
+  (extract-tangent [this tag mode]
     "If `this` is perturbed, return the tangent component paired with the
-    supplied tag. Else, returns `([[emmy.value/zero-like]] this)`."))
+    supplied tag. Else, returns `([[emmy.value/zero-like]] this)`.")
+
+  (extract-id [this id]))
+
+(defrecord Completed [v->partial]
```
sritchie, line 347: annoying that this now has to live in `differential`. It has to be here so that a `Dual` can return the correct default type if it's called with a different tag.

src/emmy/differential.cljc
```diff
@@ -352,12 +379,18 @@
...
+      (->Completed {})))
 
   #?(:clj Object :cljs default)
   (perturbed? [_] false)
   (replace-tag [this _ _] this)
-  (extract-tangent [this _] (g/zero-like this)))
+  (extract-tangent [this _ mode]
+    (if (= mode ::dual)
+      (g/zero-like this)
+      (->Completed {}))))
```
sritchie, line 393: TODO bind a singleton of `(->Completed {})` to something

src/emmy/series.cljc
```diff
@@ -50,128 +51,128 @@
   (toString [S] (str (g/freeze S)))
 
   #?@
-   (:clj
-    [IObj
-     (meta [_] m)
-     (withMeta [_ meta] (Series. xs meta))
+  (:clj
```
sritchie, line 54: formatting change that snuck in

src/emmy/tape.cljc
```diff
@@ -142,7 +140,12 @@
   ;; This implementation is called if a tape ever makes it out of
   ;; forward-mode-differentiated function. If this happens, a [[TapeCell]]
```
sritchie, line 141: TODO update the docs here. We need a better narrative now that hints at reverse-mode in the `differential.cljc` exposition.

test/emmy/tape_test.cljc
```diff
@@ -220,10 +220,11 @@
 
       (checking "d/extract-tangent" 100 [tag  gen/nat
                                          tape (sg/tapecell gen/symbol)]
-                (is (zero? (d/extract-tangent tape tag))
+                ;; TODO fix these for non-dual
```
sritchie, line 223: TODO add a test for the case where we pass a different mode

PR #170, opened 2024-04-24
Change default derivative impl to reverse mode

sritchie:
- #170:

  - changes `D` and `partial` to use reverse mode automatic differentiation by
    default, and fixes all associated tests

  - adds `emmy.generic/{zero?,one?,identity?}` implementations (all false) to
    `emmy.tape/Completed`, in case some collection type tries to simplify these
    during reverse-mode AD


src/emmy/mechanics/hamilton.cljc
```diff
@@ -224,10 +224,9 @@
                          (g/zero?
                           (g/simplify
                            (g/determinant M))))
-                  (do (println "determinant" (g/determinant M))
-                      (throw
-                       (ex-info "Legendre Transform Failure: determinant = 0"
-                                {:F F :w w})))
+                  (throw
```
sritchie, line 227: removed extra println

src/emmy/mechanics/hamilton.cljc
```diff
@@ -442,11 +441,12 @@
   "p.326"
   [C]
   (fn [s]
-    ((- J-func
-        (f/compose (Phi ((D C) s))
-                   J-func
-                   (Phi* ((D C) s))))
-     (s/compatible-shape s))))
+    (let [s-syms (s/compatible-shape s)]
```
sritchie, line 444: this forces `compatible-shape` to evaluate and make its gensym calls before the `((D C) s)` calls, important for repeatable tests

PR #172, opened 2024-05-15
Fix broken ns reference in defgeneric

yuhan0:
The current implementation relies on the macroexpansion site having `emmy.value` aliased to `v`.

I suppose we could also explicitly require emmy.value in this namespace, which lets the syntax-quote macro properly substitute it, but there could be circular dependencies I'm not aware of.


yuhan0:
There's also `(:require [emmy.value :as-alias v])`, which was only introduced in Clojure 1.11

sritchie:
Phew, thank you, this is a great catch!!

PR #175, opened 2024-06-01
Const speedups, fix Brent initial conditions

sritchie:
This PR:

  - Adds `^:const` metadata to all constants, reaping small performance wins

  - Updates
    `emmy.numerical.unimin.brent/{brent-min,brent-max,brent-min-commons,brent-max-commons}`
    to:

    - take a new `:initial-guess` argument, useful if you have some idea of
      where the minimum might lie

    - swaps the relative and absolute threshold defaults to match those from
      `scmutils`

    - adjusts the initial guess from the midpoint between `a` and `b` to a
      golden section cut (closer to `a`), to match `scmutils`

### DIFF
--- a/src/emmy/numerical/unimin/brent.cljc
+++ b/src/emmy/numerical/unimin/brent.cljc
@@ -147,6 +147,11 @@
         (or (<= fnew fx2) (= xx2 xx) (= xx2 xx1)) [[xnew fnew] x1]
         :else [x2 x1]))
 
+(def ^:private ^:const brent-error 1.0e-11)
+
+(defn- initial-brent-guess [a b]
+  (+ a (* ug/inv-phi2 (- b a))))
+
 (defn brent-min
   "Find the minimum of the function f: R -> R in the interval [a,b] using Brent's
   Method, described by Richard Brent in [Algorithms for Minimization without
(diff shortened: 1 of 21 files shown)


src/emmy/numerical/unimin/brent.cljc
```diff
@@ -183,17 +189,19 @@
   ([f a b] (brent-min f a b {}))
   ([f a b {:keys [relative-threshold
                   absolute-threshold
+                  initial-guess
                   maxiter
                   maxfun
                   callback]
-           :or {relative-threshold (g/sqrt u/machine-epsilon)
-                absolute-threshold 1.0e-11
+           :or {relative-threshold brent-error
```
sritchie, line 196: I had flipped these relative to scmutils! Flipping them back is the right call imo even though it changes some test tolerances.

sritchie:
@littleredcomputer I'm going to go ahead and merge this so we can get the bug fix out... let me know if you see any issues on review that you'd like me to tackle.

sritchie:
@alexgian this is now merged! I'll have to do a little more work to get this out into the `emmy-viewers` bundle and make it available on Maria / in the viewers... but you can give it a try via the Clerk route now if you use Emmy as a git dependency.

PR #178, opened 2024-06-14
Local extrema in univariate minimization

alexgian:
These are provided as translations of the equivalent scmutils code.
They are not integrated into the minimizer yet as options, they are simply made available for future use.


alexgian:
This was a response to #114

PR #183, opened 2024-08-08
Add extract-id implementations for all types

sritchie:
- #183:

  - adds `emmy.{autodiff, tape}` to `emmy.sci`'s exported namespace set

  - adds `emmy.dual/extract-id` implementations for all supported output types
    (every type that already implements `emmy.dual/IPerturbed`)

  - moves `emmy.tape/Completed` to `emmy.dual/Completed`; it doesn't really make
    sense there, but this is the only current way to remove the circular
    dependency between `emmy.dual` and `emmy.tape`. (`tape` needs a `dual`
    import to gain `emmy.dual/IPerturbed`.)

  - simplifies the `emmy.tape/gradient` implementation down to only handle
    single real-or-structural arguments, just like `emmy.dual/derivative`. We'll
    share the "handle-multiple-input" implementation between the two in a
    follow-up PR

  - makes the tests in `emmy.calculus.derivative` generic on the derivative
    implementation, so we can run all tests in forward and reverse mode.

### DIFF
--- a/src/emmy/collection.cljc
+++ b/src/emmy/collection.cljc
@@ -53,7 +53,7 @@
   f/IArity
   (arity [_] [:between 1 2])
 
-  ;; [[d/replace-tag]] and [[d/extract-tangent]] pass the buck down the vector's
+  ;; The [[emmy.dual/IPerturbed]] functions pass the buck down the vector's
   ;; elements.
   d/IPerturbed
   (replace-tag [v old new] (mapv #(d/replace-tag % old new) v))
--- a/src/emmy/collection.cljc
+++ b/src/emmy/collection.cljc
@@ -53,12 +53,12 @@
   f/IArity
   (arity [_] [:between 1 2])
 
-  ;; Vectors are functors, so they can be perturbed if any of their elements are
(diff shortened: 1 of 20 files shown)


src/emmy/collection.cljc
```diff
@@ -53,12 +53,12 @@
...
-  ;; Vectors are functors, so they can be perturbed if any of their elements are
-  ;; perturbed. [[d/replace-tag]] and [[d/extract-tangent]] pass the buck down
-  ;; the vector's elements.
+  ;; [[d/replace-tag]] and [[d/extract-tangent]] pass the buck down the vector's
+  ;; elements.
   d/IPerturbed
   (replace-tag [v old new] (mapv #(d/replace-tag % old new) v))
-  (extract-tangent [v tag mode] (mapv #(d/extract-tangent % tag mode) v)))
+  (extract-tangent [v tag mode] (mapv #(d/extract-tangent % tag mode) v))
+  (extract-id [v id] (mapv #(d/extract-id % id) v)))
```
sritchie, line 61: these tests will get hit when I enable gradients in the next PR.

PR #185, opened 2024-08-12
add forward, reverse-mode AD operators, enable tests for reverse-mode

sritchie:
- adds a dynamic variable `emmy.calculus.derivative/*mode*` that allows the
    user to switch between forward and reverse mode automatic differentiation

  - adds a new `emmy.calculus.derivative/gradient` that acts like
    `emmy.tape/gradient` but is capable of taking multiple variables

  - adds new operators `emmy.calculus.derivative/{D-forward, D-reverse}` and
    operator-returning-functions `emmy.calculus.derivative/{partial-forward,
    partial-reverse}` that allow the user to explicitly invoke forward-mode or
    reverse-mode automatic differentiation. `D` and `partial` still default to
    forward-mode

  - modifies `emmy.tape/gradient` to correctly error when passed invalid
    selectors, just like `emmy.dual/derivative`.

### DIFF
--- a/src/emmy/calculus/derivative.cljc
+++ b/src/emmy/calculus/derivative.cljc
@@ -143,7 +143,7 @@
               (str "Selectors " selectors
                    " not allowed for non-structural input " input)))))))
 
-(defn multi
+(defn- multi
   "Given
 
     - some higher-order function `op` that transforms a function of a single
--- a/src/emmy/calculus/derivative.cljc
+++ b/src/emmy/calculus/derivative.cljc
@@ -36,7 +36,8 @@
 ;; To generate the result:
 ;;
 ;; - For a single non-structural argument, return `(d/derivative f)`
-;; - else, bundle up all arguments into a single [[s/Structure]] instance `xs`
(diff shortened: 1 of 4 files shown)


src/emmy/calculus/derivative.cljc
```diff
@@ -192,24 +229,66 @@
...
+;; reverse-mode AD.
+;;
 ;; TODO: I think this is going to cause problems for, say, a Structure of
 ;; PowerSeries, where there is actually a cheap `g/partial-derivative`
 ;; implementation for the components. I vote to back out this `::s/structure`
 ;; installation.
 
+(def ^:dynamic *mode* d/FORWARD-MODE)
+
 (doseq [t [::v/function ::s/structure]]
```
littleredcomputer, line 242: The dynamic variable bothers me a tiny bit so I want to see if it's really necessary. I suspect it is, but here's one question. I can see having D-forward and D-reverse, and letting D be bound to one or the other by default. That would be enough, except we need to know which D partial differentiation implies, and we don't want the naming to mushroom into having two options for everything differentiation-like. But isn't there a clear choice when you want just one partial derivative out of many?

